PR #160, opened 2015-01-06
added skipTests flag support

lkwg82:
I would like the have the 'skipTests' flag as well.

Here is my PR ;)

Cheers Lars


hcoles:
thanks!

PR #167, opened 2015-01-23
Added a configuration element to allow classpath elements to be excluded

TomRK1089:
Added a configuration element to allow classpath elements to be excluded from mutation testing, for the case when those elements might be troublesome. This mirrors the similar options in Surefire and Failsafe. Added a corresponding unit test to exercise the new config option.

I will be honest, I have not developed a Maven plugin before, so there may be a more efficient/idiomatic way to remove artifacts from the classpath being built up in the MojoToReportOptionsConverter -- so any suggestions or corrections there are welcome. :)

### DIFF
--- a/pitest-maven/src/main/java/org/pitest/maven/MojoToReportOptionsConverter.java
+++ b/pitest-maven/src/main/java/org/pitest/maven/MojoToReportOptionsConverter.java
@@ -55,6 +55,14 @@ public class MojoToReportOptionsConverter {
     addOwnDependenciesToClassPath(classPath);
 
     classPath.addAll(this.mojo.getAdditionalClasspathElements());
+    
+    for(Object artifact : this.mojo.getProject().getArtifacts()) {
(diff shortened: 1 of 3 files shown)


Milstein:
This can be more troublesome if there are lots of classes to be removed! How about a configuration element to allow mutation of a particular classpath/ paths (Comma separated) among large pool of paths.

TomRK1089:
@Milstein , this element would let you remove an entire artifact, the same way Surefire and Failsafe do:
```
<classpathDependencyExcludes>
  <param>com.google:android</param>
</classpathDependencyExcludes>
```

rather than a single class at a time. This prevents classes in those artifacts not only from being mutated but loaded at all, which is great when there are potential conflicts from library authors (cough cough, Android core team) who have rebundled classes impromperly.

hcoles:
@TomRK1089 - thanks for this. It will go out in the 1.1.4 release in the next few days.

PR #180, opened 2015-03-14
[#179] Fix problem with using launch classpath

### LINKED ISSUE #179: Broken `includeLaunchClasspath=false` on Windows
Issue originally reported to [gradle-pitest-plugin (#26)](https://github.com/szpak/gradle-pitest-plugin/issues/26). After code analyze it turned out to be a problem in PIT with parsing and comparing launch classpath.

The problem is that with `includeLaunchClasspath=false` classpath elements are taken with `ClassPath.getClassPathElements()` which returns elements as were provided in `-cp` (e.g. `c:\path\pitest-1.1.5.jar`). It is later compared in `PluginFilter` with `getClass().getProtectionDomain().getCodeSource().getLocation().getFile()` which returns path from URI (e.g. `/C:/path/pitest-1.1.5.jar`). A comparison of strings fails and `pitest.jar` is not on a classpath which ends on Windows with:

```
Exception in thread "main" org.pitest.util.PitError:
Unable to load class content for org.pitest.boot.HotSwapAgent
```

In addition there could be problem also on Linux where symlinks are used or launch classpath elements are passed as relative paths. One of the solutions is to compare paths as canonical paths.

szpak:
Paths are compared as canonical paths to prevent comparison problem on Windows (and in some cases also on Linux). Fixes #179.

Unfortunately code became less readable, but collection mapping isn't compact in Java (especially <8 and especialy with checked exceptions). I had problem with testing it automatically and end up creating quite hacky integration test which doesn't match other unit tests in `OptionsParserTest` (should be moved to a separate file?). Maybe you have a better idea how to test it in a reliable way.

It is possible that similar problem with paths appears also in some fragments of code.

### DIFF
--- a/pitest-command-line/src/main/java/org/pitest/mutationtest/commandline/OptionsParser.java
+++ b/pitest-command-line/src/main/java/org/pitest/mutationtest/commandline/OptionsParser.java
@@ -63,7 +63,6 @@ import joptsimple.OptionSpec;
 import joptsimple.OptionSpecBuilder;
 
 import org.pitest.classpath.ClassPath;
-import org.pitest.functional.FArray;
 import org.pitest.functional.FCollection;
 import org.pitest.functional.predicate.Predicate;
 import org.pitest.mutationtest.config.ConfigOption;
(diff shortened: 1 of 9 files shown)


hcoles:
Merged in last night - I'll put a release out soon since this is blocking people.

Thanks.

Issue #10, opened 2014-01-07
Add maven goal for PIT report generation and maven site integration


hcoles:
_From [stefan.p...@googlemail.com](https://code.google.com/u/115492280750533935957/) on September 18, 2011 13:45:09_

PIT is currently hard coded to produce only a html report (see Issue 9 ). After revising the reporting stuff an enhanced maven plugin would be nice that generates a report integrated into the maven site.

_Original issue: http://code.google.com/p/pitestrunner/issues/detail?id=10_

dustinbarnes:
Is there any plan to get this in there? I'm currently trying to hack around it, but this would be the thing that really cleanly integrates into the code quality lifecycle.

hcoles:
@dustinbarnes Although I agree this would be good to have, I'm afraid it is unlikely to happen unless some-one jumps in and contributes a fix.

dustinbarnes:
That sounds fair enough. It's not so much a fix as a new feature request, really. 

If I have some time, I might take a whack at it. I've created some maven mojo plugins, but never a reporting plugin. People can always go modify a site.xml and dump the PIT stuff into the ${basedir}/target/site directory, but it's kind of a pain.

hcoles:
Pull requests gratefully received.

jasonmfehr:
I'm working on writing a reporting mojo.  I'm not very far along, and given my current schedule, it will be a few weeks.  If you need it quicker and are going to work on it, please let me know so I can stop working on mine.

lkwg82:
I'm curious about the status.

jasonmfehr:
The initial development is done (other than fixing the long running integration tests).  In its current state, the reporting mojo simply copies the contents of the target/pit-reports directory.

I really want to add the ability to use the xml or csv files along with the doxia sink to generate a report that is fully integrated into the maven site, but that is not something I will have time to work on for a couple months.  If anyone else wants to work on it, I would be more than willing to help out where I can.

lkwg82:
Maybe this issue can be split into `done` and `to do`? An issue with commits and status `unfinished` for long time looks like some blocker. 

This makes the status and progress much more clearer.

jasonmfehr:
Yeah, definitely.  Want me to take care of closing down this issue an opening a new one?

lkwg82:
Would be the best. You are on the topic and you know what have done and what is open. This also supports the maintainer to make releases and release notes. Thx.

jasonmfehr:
[...]

PR #187, opened 2015-04-19
Add Report Goal

jasonmfehr:
This pull request creates a maven reporting goal for the pitest report.  Issue https://github.com/hcoles/pitest/issues/10 documents the request to create a reporting goal for the maven plugin.

Since the existing maven plugin automatically generates an html report, the approach I took was to copy the files comprising that latest html report (if multiple reports have been generated with the timestampedReports parameter set to true) into the maven site directory.  I felt this approach was the least intrusive while maintaining full backwards compatibility.

The report mojo and its supporting classes are all contained within the org.pitest.maven.report package and its subpackages in the pitest-maven project.  The pitest-maven-verification project has several new tests to verify the correct HTML report is copied to the correct location.  All the test directories start with "pit-site".


jasonmfehr:
My bad on the failing CI build.  Fixed it by changing the default of the siteReportDirectory default.  Also updated the integration tests to have better asserts.

pitest-maven/src/main/java/org/pitest/maven/report/PitReportMojo.java
```diff
@@ -0,0 +1,190 @@
...
+        super();
+        
+        this.reportGenerationManager = new ReportGenerationManager();
+    }
+    
+    public String getOutputName() {
+        return this.siteReportDirectory + File.separator + "index";
+    }
+
+    public String getName(Locale locale) {
```
KyleRogers: @hcoles as far as I know pitest needs Java 6 for execution. Right? We should add '''@Override''' here if no Java 5 source compatiblity is required. Same applies for '''getOutputName()''' and '''getDescription(...)'''

Took me some time to figure out that those methods implement some methods from MavenReport interface and will be called by maven to build the navigation.

pitest-maven/src/main/java/org/pitest/maven/report/generator/IReportGenerationStrategy.java
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+package org.pitest.maven.report.generator;
+
+
+
+public interface IReportGenerationStrategy {
```
KyleRogers: pitest doesn't prefix interfaces with capital I . I think we'd better name it ReportGenerationStrategy.

pitest-maven/src/main/java/org/pitest/maven/report/generator/XMLReportGenerator.java
```diff
@@ -0,0 +1,33 @@
...
+ * 
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+package org.pitest.maven.report.generator;
+
+
+public class XMLReportGenerator implements IReportGenerationStrategy {
```
KyleRogers: Maybe we should just drop that class? Whats the purpose of having a non-implemented report generator? Or do you plan to implement it later?

pitest-maven/src/main/java/org/pitest/maven/report/generator/ReportGenerationManager.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+import org.pitest.maven.report.ReportSourceLocator;
+import org.pitest.util.PitError;
+
+public class ReportGenerationManager {
+
+    protected ReportSourceLocator reportLocator;
+    protected List<IReportGenerationStrategy> reportGenerationStrategyList;
```
KyleRogers: I'm not a fan of too broad visibility for testing puposes only. Maybe you could add a second constructor that allows injection of resource locator and strategy list?
Afterwards you can make those properties private and final.

The no-arg constructor will be a kind of factory method with this approach. You could also factor out an explicit, static factory method, if you like.

KyleRogers:
Good job jason, I just have some minor comments regarding code style ;-).

@hcoles  what do you think?

hcoles:
@jasonmfehr Thanks for this - it looks good, and thanks @KyleRogers for looking through this.

I'll try and get this merged in this week and fix any minor stylistic stuff during the merge.

jasonmfehr:
Thanks @KyleRogers for the feedback.  @hcoles please give me a couple days to go through the feedback and get those changes made before merging.

pitest-maven/src/main/java/org/pitest/maven/report/generator/IReportGenerationStrategy.java
jasonmfehr: Good catch.  I meant to change that name to remove the leading "I" anyway but forgot.

pitest-maven/src/main/java/org/pitest/maven/report/generator/XMLReportGenerator.java
jasonmfehr: By copying over the pit html report, the maven site opens it in a new window instead of within the context of the maven site (i.e. so the site header and left-side navigation are not visible).  I was hoping to to use the doxia sink to generate a more integrated report from the pit xml data file but did not have the time right now.  My line of thought was to build it later or have someone else build it.

I will definitely drop the class if that is the better design.  It would be very easy to recreate later.

pitest-maven/src/main/java/org/pitest/maven/report/generator/ReportGenerationManager.java
jasonmfehr: I'm not a fan either of having to write code a certain way just for testing.  It's usually an indicator of a bad design.  Great idea using a second constructor.  I am definitely going to do that.

jasonmfehr:
I committed the fixed from the code review.  Please take another look @KyleRogers.  I also discovered one of the integration tests was brittle.  That is fixed as well.

jasonmfehr:
Also, I will send another pull request in a week or two with updates to the pitest-site for the new reporting mojo.  Please let me know if you need the updates sooner.

hcoles:
I had a go at merging this last night but was blocked by a failing test. It eventually ran green without a change - don't have the failure trace to hand but will post it this evening if I can reproduce.

jasonmfehr:
Please let me know if you see that test fail again so I can dig into what is happening.  I originally had a few issues with brittle integration tests but fixed all of them that I saw.

PR #189, opened 2015-04-28
PitMojo.java - make execute() non-final

tclausing:
I want to implement a new MutationResultListener, but it needs additional configuration which the api doesn't provide for. So my intention is to extend PitMojo, add my extra params, store them statically, and use them in my MutationResultListener.

### DIFF
--- a/pitest-maven/src/main/java/org/pitest/maven/PitMojo.java
+++ b/pitest-maven/src/main/java/org/pitest/maven/PitMojo.java
@@ -332,7 +332,7 @@ public class PitMojo extends AbstractMojo {
     this.plugins = plugins;
   }
 
-  public final void execute() throws MojoExecutionException,
+  public void execute() throws MojoExecutionException,
   MojoFailureException {
 
     switchLogging();


hcoles:
Creating a subclass of the PitMojo for each custom listener is not a scalable solution for passing configuration around.

A proper mechanism is needed to pass free form data to plugins - it ought to be fairly straightforward to update the current plugin interface to take a Properties object.

tclausing:
To clarify, I need to add dynamic config by way of maven's @parameter.  To do that without duplicating yours, afaik I need to extend PitMojo.  Also, I intend to add my listener as a plugin dependency rather than to the pitest artifact itself, so while I agree it's not scalable, it won't impact your artifact, and for my own use I'm not real concerned about it.

tclausing:
I'm sorry, that probably didn't make much sense. Normally I'd add this stuff in a fork of pitest-maven, but to get that through my org's red tape would not be pleasant, and I'm also hoping to avoid the overhead of keeping up to date with an upstream.

hcoles:
I've just knocked something together locally for #190 so a map of free-form properties can be passed to the plugin from the mojo. I assume this would solve your problem?

tclausing:
Don't the extra params need to be explicit and present at build time so that the plugin descriptor can capture them?

hcoles:
@tclausing Free form key value pairs can be captured in a map.

eg

```
<pluginProperties>
   <foo>fish</foo>
   <bar>cthulhu</bar>
</pluginProperties>
```

PR #191, opened 2015-04-29
Minor Fixes for PIT Report Mojo

jasonmfehr:
While writing updates to the pit documentation, I noticed one of the user properties was wrong.  I updated the ``sourceDataFormats`` user property from ``${sourceDataFormats}`` to ``${pit.report.sourceDataFormats}``.  The original value was left over from before I settled on a naming strategy for the user properties.  The new value matches the convention for the reporting mojo's user properties to all start with "pit.report."

I added an integration test that verifies the build fails if the reporting mojo is executed without also executing the ``mutationCoverage`` goal.

I also noticed a build had failed because the integration tests ran too long without writing any output.  I added the travis_wait function to the .travis.yml file.  This addition causes a line to be outputted every  minute or so for 20 minutes thus increasing the build timeout from 10 minutes to 20 minutes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,4 +3,4 @@ jdk:
   - openjdk6
   - oraclejdk7
   - oraclejdk8 
-script: travis_wait mvn clean verify
+script: mvn clean verify
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,4 +3,4 @@ jdk:
   - openjdk6
   - oraclejdk7
   - oraclejdk8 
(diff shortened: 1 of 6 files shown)


jasonmfehr:
Apparently travis_wait does not do quite what the documentation says it does.  When the initial build ran, it paused for 10 minutes then timed out.  I removed the travis_wait from the .travis.yml file and the build ran green.  Later I may look into parallelizing the integration tests so they run faster.

Issue #190, opened 2015-04-28
Provide mechanism for configuring custom plugins


hcoles:
Custom plugins are can currently only be configured by the built in properties, or by environment variables.

A mechanism is required whereby they can be passed free form options e.g as a Java properties object.

hcoles:
All tool classpath plugins now accept a Properties object to their factory method. Custom properties can now be set within a nested configuration section of the pitest config via maven.

```xml
<plugin>
    <groupId>org.pitest</groupId>
    <artifactId>pitest-maven</artifactId>
    <version>LATEST</version>
    <configuration>
        <targetClasses>
            <param>com.your.package.root.want.to.mutate*</param>
        </targetClasses>
        <configuration>
           <foo>valueneededbyplugin</foo>
        </configuration>
    </configuration>
</plugin>
```

Open questions

1) Is there any issue with also allowing this for client classpath plugins? (probably not)
2) Is "configuration" a good choice of name for the nested tag?
3) How should these properties be set via ant/commandline/gradle?

@szpak Any thoughts/preferences for how this is handled? Is the gradle plugin still implemented in terms of the command line tool?

KyleRogers:
Let me comment on your open questions:
1) depending on the way configuration is sent to the client (as system properties?) and the amount of configuration there was a problem with older windows systems. Don't know if Windows 7 and above suffer the same limitations but surefire users had problems with too long classpaths on windows systems.
2) Maybe the tag could be named ```<pluginConfiguration>``` to emphasize, that this tag contains configuration for plugins only and to avoid confusion with maven's ```<configuration>``` Tag
3) for command line I would suggest:
```
java -cp <your classpath including pit jar and dependencies> \
    org.pitest.mutationtest.commandline.MutationCoverageReport \
    --reportDir  \
    --targetClasses com.your.package.tobemutated* \
    --targetTests com.your.packge.*
    --sourceDirs  \
    --pluginConfiguration "key1=value" \
    --pluginConfiguration "key2=value" \
```
where the first = will be the separator for key and value (thus = may not be used in configuration property keys)

hcoles:
@KyleRogers 

Configuration would most likely be serialised as part of the existing SlaveArguments class which shouldn't cause any OS specific problems.

`pluginConfiguration` is certainly a better name. I was toying with the idea of also using for non plugin configuration (given the pain involved in adding a new parameter to all the various build integrations), but this is probably a bad idea.

[...]

PR #192, opened 2015-05-02
Adding Maven Mojo PluginProperties (java.util.Properties) parameter

willhuang85:
Provide mechanism for configuring custom plugins #190


hcoles:
Thanks for this contribution, it is appreciated. I can't generally accept changes without tests of the behaviour they add. Fortunately I had a change very similar to this that I'd made locally which has now been merged in.

PR #202, opened 2015-05-13
changed travis file to test against multiple versions of maven and jdk

lkwg82:
this results in #(jdk-versions) x #(maven-version) = 3 x 4 = 12 runs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,24 @@
 language: java
+
 jdk:
   - openjdk6
   - oraclejdk7
-  - oraclejdk8 
-script: mvn clean verify
+  - oraclejdk8
+
+env:
+  global:
+  # taken first download mirror from http://maven.apache.org/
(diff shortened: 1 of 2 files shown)


lkwg82:
Building with maven 3 and testing also with maven 2 during build is kind of tricky. Again the question is the the effort worth?

hcoles:
The building and testing are currently pushed together but ideally wouldn't be.

It only needs to build on maven 3. It would be nice to test the pitest-maven plugin on multiple mavens, and to do so without rebuilding each time.

The latest snapshot is always tested against maven 2 is tested over here -> https://hjc.ci.cloudbees.com/job/maven2_triangle_example/

This currently only happens after a snapshot is manually published, and not at all for PRs. It would make sense to auto-deploy snapshots after a merge into the mainline so feeback is obtained a little faster.

People are still using maven 2 and I'll continue to support it for a while longer.

PR #204, opened 2015-05-17
Cleanup scm mojo while debugging

lkwg82:


pitest-maven/src/main/java/org/pitest/maven/PitMojo.java
```diff
@@ -47,7 +47,7 @@
    * @parameter expression="${targetClasses}"
    *
    */
-  protected ArrayList<String>         targetClasses;
+  private List<String>         targetClasses;
```
hcoles: Unfortunately maven 2 breaks unless given a concrete list type

lkwg82: I know and you already merged upgrade to maven 3, see #199 

So this should be not an issue.

hcoles: I'd assumed this just enforced the minimum version for building pitest. Does it set a minimum for someone running the built plugin?

lkwg82: I think we did not talk about this specific expectation. I'd assume both.

But maybe we can be clear on #202 to meet our both expactations ;).

hcoles: Indeed - was planning on merging that in when I'd had chance to look properly at the very slow site report tests. If you could update the PR to add and ignore to them I'll click the merge button now.

lkwg82:
Which pr? This or #202?

hcoles:
@lkwg82 #202

hcoles:
The less immediate feedback from https://hjc.ci.cloudbees.com/job/maven2_triangle_example/configure confirms that maven 2 is still working fine on master. Change to interface types will break it.

PR #207, opened 2015-05-18
Cleanup pitmojo while debugging

lkwg82:


hcoles:
@lkwg82 If you look in the root of the project you'll find eclipse format and code cleanup profiles - these describe the style of code in use in the project. They were written a long time ago so I don't necessarily represent my current views on code style, but for consistency additions to the codebase should stick to them rather than using their personal style.

My current view is that class members should always be final unless there is a good reason for them not to be. Any variable within a method that has scope within multiple blocks should be final unless it is intended to be overwritten (which it almost always shouldn't). Method parameters need not be final unless the method is more than a few lines long.

From memory the cleanup profile is a bit too final adding happy. It should probably be updated to at least stop adding them for parameters.

Some people like to prepend this, some don't. I do it simply because it helps with autocomplete in the IDE - the profile always prepends this so to keep things consistent.

lkwg82:
> My current view is that class members should always be final unless there is a good reason for them not to be.

I fully agree.

> Any variable within a method that has scope within multiple blocks should be final unless it is intended to be overwritten (which it almost always shouldn't).

I think with clean code we overcome big blocks and oversee the reassignments. The code should be expressive in a way and those bad practices should be discussed in core review, clean and not redundant, except it is needed to decrease complexity.

> Method parameters need not be final unless the method is more than a few lines long.
This is a workaround for bad practice to reassign parameters and long methods. Both we should change and demand in code reviews. I'd like to change this rule, if you agree.

> From memory the cleanup profile is a bit too final adding happy. 
Here is one aspect of memory usage discussed and it says `final` has no special impact:
http://stackoverflow.com/questions/306862/does-using-final-for-variables-in-java-improve-garbage-collection

>It should probably be updated to at least stop adding them for parameters.
I dont have eclipse. Will you do that?

> Some people like to prepend this, some don't. I do it simply because it helps with autocomplete in the IDE - the profile always prepends this so to keep things consistent.

Is this another workaround for bad practice of 'too many lines' you dont oversee where the code is from? If we need to many hints in our code maybe it is too complex. What about making it easier. And in the end it is also of course it a question of style. ;)

lkwg82:
> Some people like to prepend this, some don't. I do it simply because it helps with autocomplete in the IDE - the profile always prepends this so to keep things consistent.

If the 'this' is ommitted, there are three valid locations the variable could be found:
- as a local variable
- as a parameter
- as a member of `this` class

In the case the variable is in a parent class, this is a violation and should be changed.

In my opinion 'this' is boilerplate code and can ommited if you practice clean code.

PR #208, opened 2015-05-26
Fix slow report mojo tests

jasonmfehr:
Fixed the slow running integration tests of the reporting mojo in PitMojoIT.  The fix is to set up the test as if a mvn clean install had been executed.  Then, only the site goal is executed.


pitest-maven-verification/src/test/java/org/pitest/PitMojoIT.java
```diff
@@ -398,32 +323,73 @@ private File buildFile(File base, String... pathParts) {
 	  StringBuilder path = new StringBuilder(base.getAbsolutePath());
 
 	  for (String part : pathParts) {
-		  path.append(File.separator).append(part);
+	      if(!StringUtils.isBlank(part)) {
```
lkwg82: Would be an IllegalArgumentException more appropriatly? I think any blank is an illegal part of a path. Except I'd like to   hide sth.

pitest-maven-verification/src/test/java/org/pitest/PitMojoIT.java
```diff
@@ -149,13 +157,15 @@ public void shouldCorrectlyTargetTestsWhenMultipleBlocksIncludeALine() throws Ex
...
-
-    verifier.executeGoals(Arrays.asList("test", "org.pitest:pitest-maven:mutationCoverage", "site"));
-
+    File testDir = prepareSiteTest("/pit-site-skip");
     File siteParentDir = buildFile(testDir, "target", "site");
+    
+    verifier.executeGoal("site");
+
+    assertThat(siteParentDir.exists()).isEqualTo(true);
     assertThat(buildFile(siteParentDir, "pit-reports").exists()).isEqualTo(false);
```
lkwg82: `assertThat` has a FileAssert-extension, so it is sufficient to write `assertThat(...).exists()`

same with `assertThat(...).isTrue()`

pitest-maven-verification/src/test/java/org/pitest/PitMojoIT.java
```diff
@@ -149,13 +157,15 @@ public void shouldCorrectlyTargetTestsWhenMultipleBlocksIncludeALine() throws Ex
...
-    File testDir = prepare("/pit-site-skip");
-
-    verifier.executeGoals(Arrays.asList("test", "org.pitest:pitest-maven:mutationCoverage", "site"));
-
+    File testDir = prepareSiteTest("/pit-site-skip");
     File siteParentDir = buildFile(testDir, "target", "site");
+    
+    verifier.executeGoal("site");
+
+    assertThat(siteParentDir.exists()).isEqualTo(true);
```
lkwg82: `assertThat` has a FileAssert-extension, so it is sufficient to write `assertThat(...).exists()`

same with `assertThat(...).isTrue()`

pitest-maven-verification/src/test/java/org/pitest/PitMojoIT.java
```diff
@@ -398,32 +323,73 @@ private File buildFile(File base, String... pathParts) {
...
+      
+      FileUtils.copyDirectory(buildFile(testDir, "src", "test", "resources", "pit-reports"), testTargetDir);
+      
+      return testDir;
+  }
+  
+  private void verifyPitReportTest(File testDir) throws Exception {
+      File pitReportSiteDir = buildFile(testDir, "target", "site", "pit-reports");
+      
+      //assert that the expected pit reports directory exists under the site directory and that the directory contains the correct files
```
lkwg82: this comment can be omitted in favor of expressive code below (it almost is)

pitest-maven-verification/src/test/java/org/pitest/PitMojoIT.java
```diff
@@ -398,32 +323,73 @@ private File buildFile(File base, String... pathParts) {
...
+      assertThat(pitReportSiteDir.exists()).isEqualTo(true);
+      assertThat(this.buildFile(pitReportSiteDir, "marker_expected.txt").exists()).isEqualTo(true);
+      
+      //assert that the expected report name/description was written to the site project report's index.html file
```
lkwg82: please write the code this way, you dont need the comment anymore

pitest-maven-verification/src/test/java/org/pitest/PitMojoIT.java
```diff
@@ -398,32 +323,73 @@ private File buildFile(File base, String... pathParts) {
...
+   * The provided {@code testPath} must have this folder structure set up underneath it:  src/test/resources/pit-reports  
+   * The contents of this directory will be moved to the target directory simulating a mvn clean install.   
+   * 
+   * @param testPath {@link String} location of the test to set up, this path is relative to <code>${basedir}/src/test/resources</code>
+   * 
+   * @return {@link File} representing the temporary folder that was set up for this execution of the test
+   * @throws Exception
+   */
+  private File prepareSiteTest(String testPath) throws Exception {
+      File testDir = prepare(testPath); //temporary location where the test will execute
```
lkwg82: can we rename `testDir` to `temporaryTestLocation` to make the comment needless?

pitest-maven-verification/src/test/java/org/pitest/PitMojoIT.java
```diff
@@ -398,32 +323,73 @@ private File buildFile(File base, String... pathParts) {
...
+  private File prepareSiteTest(String testPath, String latestDir) throws Exception {
+    File testDir = prepareSiteTest(testPath);
+    File testTargetDir = this.buildFile(testDir, "target", "pit-reports"); //location where the target directory would be if a mvn clean install was executed
+    DirectoriesOnlyWalker walker = new DirectoriesOnlyWalker();
+    
+    for(File f : walker.locateDirectories(testTargetDir)) {
+        f.setLastModified(0L);
     }
+    
+    buildFile(testTargetDir, latestDir).setLastModified(System.currentTimeMillis());
```
lkwg82: Would you be so kind and rename the method `buildFile()` to `buildFilePath()` this describes more what it actually does. Thx!

pitest-maven-verification/src/test/java/org/pitest/PitMojoIT.java
```diff
@@ -149,13 +157,15 @@ public void shouldCorrectlyTargetTestsWhenMultipleBlocksIncludeALine() throws Ex
    */
   @Test
   public void shouldSkipSiteReportGeneration() throws Exception {
-    File testDir = prepare("/pit-site-skip");
-
-    verifier.executeGoals(Arrays.asList("test", "org.pitest:pitest-maven:mutationCoverage", "site"));
-
+    File testDir = prepareSiteTest("/pit-site-skip");
```
lkwg82: is the comment of this test really needed? I think the name is perfect and the comment redundant.

pitest-maven-verification/src/test/java/org/pitest/PitMojoIT.java
fge: `.setLastModified()` returns a boolean; check for the result!

pitest-maven-verification/src/test/java/org/pitest/support/DirectoriesOnlyWalker.java
```diff
@@ -0,0 +1,61 @@
...
+ * </code>
+ */
+public class DirectoriesOnlyWalker extends DirectoryWalker {
+
+    public DirectoriesOnlyWalker() {
+        super(FileFilterUtils.directoryFileFilter(), -1);
+    }
+    
+    public List<File> locateDirectories(File startDir) {
+        List<File> foundDirs = new LinkedList<File>();
```
fge: Why a `LinkedList` and not an `ArrayList`?

pitest-maven-verification/src/test/java/org/pitest/PitMojoIT.java
jasonmfehr: Given that this is a private helper function within test code, I think its not necessary to program that defensively.  I removed the check for isBlank and added javadoc.  I can easily throw an IllegalArgumentException if pathParts is empty or any item within pathParts is empty if you think that is warranted.

pitest-maven-verification/src/test/java/org/pitest/PitMojoIT.java
jasonmfehr: Good catch.  I cleaned up all my assertThat statements.

pitest-maven-verification/src/test/java/org/pitest/PitMojoIT.java
jasonmfehr: I tweaked the variable names and asserts.  I also removed the comments.  Please let me know if you have other ideas to make it more readable.

pitest-maven-verification/src/test/java/org/pitest/PitMojoIT.java
jasonmfehr: Done.  I used ```tempTestExecutionDir``` which I felt was a slightly better indicator of what that variable contains.

pitest-maven-verification/src/test/java/org/pitest/PitMojoIT.java
jasonmfehr: Fixed both of these.

pitest-maven-verification/src/test/java/org/pitest/PitMojoIT.java
jasonmfehr: I started by putting in comments for each test scenario I wanted to cover and then added the tests under each comment.  I left the comments in place as another form of documentation, but I think at this point I could delete all the comments.  Do you see any value in keeping them?  If not, I will remove them.  I did rename a few tests to better clarify what they are testing.

pitest-maven-verification/src/test/java/org/pitest/support/DirectoriesOnlyWalker.java
jasonmfehr: I used ```LinkedList``` because the returned list would definitely have lots of inserts and most likely very few retrieves outside an ```Iterator```.  That being said, the returned ```List``` will have very few entries so either ```ArrayList``` or ```LinkedList``` will work just fine.  If convention supports using ```ArrayList```, I will gladly change it.  Please let me know.

pitest-maven-verification/src/test/java/org/pitest/PitMojoIT.java
jasonmfehr: I think I made the updates you suggested, but please double check and let me know if I missed any.

PR #211, opened 2015-06-11
fixed bug with an interface as a closeable (try-with-resources statement)

artspb:
JVM uses different instructions to invoke methods on an interface (INVOKEINTERFACE) and a class (INVOKEVIRTUAL). The previous implementation was considering only the second case. Such a behavior was leading to false positives on the auto-generated finally block in case of an interface as a closeable. Example 7 demonstrates the appropriate situation: https://github.com/artspb/try-with-resources-examples


artspb:
Could anybody give me a hint what's wrong with the build? I can't really find a reason...

lkwg82:
It seems to me like a (temporary) network failure. The build breaks at downloading a dependency.

```
Could not transfer artifact joda-time:joda-time:jar:1.6.2 from/to central (https://repo.maven.apache.org/maven2): GET request of: joda-time/joda-time/1.6.2/joda-time-1.6.2.jar from central failed: Connection reset -> [Help 1]
```

Just rerun the build and it might go away.

hcoles:
@artspb The change looks like it might risk triggering for non try-with-resource statement bytecode sequences.

The original code made a check that the call was to a "close" method. This looks to have been removed. I'd imagine the chances of the compiler generated the same 14 opcode sequence for non try-with-resources code is high enough that it will happen.

Can the check be made stronger to ensure that there is a call to "close" in the sequence, or there is a java.lang.AutoCloseable on the stack?

artspb:
@hcoles I've removed this check because from my point of view it's quite unlikely to have the 14 (or even 17) instructions in the certain order **on the same line**. The check can be implemented, but it'll look ugly (because of complicated if-else statement) and little bit slow down performance (because of String comparisons). I would prefer to leave it as is. Do you really sure that it's needed?

hcoles:
@artspb I don't think performance from string comparisons is likely to be a major concern, but as the sequence must occur on a single line I agree false triggers are unlikely to be an issue. 

Merging now, thanks.

PR #214, opened 2015-07-05
Small changes for HTML report generator

akhayrutdinov:
- set monospaced font-family for sources' listings
- cache CSS when writing html report

@hcoles, do mind reviewing it? I believe monospaced text is better for displaying sources.
(the previous PR was send from my old account, I've deleted the account and closed the PR)

### DIFF
--- a/pitest/src/main/resources/templates/mutation/style.css
+++ b/pitest/src/main/resources/templates/mutation/style.css
@@ -18,10 +18,11 @@ body{
 }
 
 .src { 
-border: 1px solid #dddddd;
-padding-top: 10px;
-padding-right: 5px;
-padding-left: 5px;
+	border: 1px solid #dddddd;
+	padding-top: 10px;
+	padding-right: 5px;
+	padding-left: 5px;
+	font-family: Consolas, Courier, monospace;
 }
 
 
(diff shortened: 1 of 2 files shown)


hcoles:
I agree the monospaced text is a good idea - but it looks little small by default to me.

Does caching the CSS make a significant difference to performance? The mutatable field is a little ugly - could the css not just be resolved once and passed in by the constructor?

Thanks

Henry

pitest/src/main/resources/templates/mutation/style.css
```diff
@@ -18,10 +18,12 @@ body{
...
-border: 1px solid #dddddd;
-padding-top: 10px;
-padding-right: 5px;
-padding-left: 5px;
+	border: 1px solid #dddddd;
+	padding-top: 10px;
+	padding-right: 5px;
+	padding-left: 5px;
+	font-family: Consolas, Courier, monospace;
+	font-size: 12px;
```
lkwg82: what about leave the size it to its defaults (looks ok for me)

lkwg82:
@akhayrutdinov I agree with @hcoles to putting `css` as a only once read to the construtor or even as a static field (but I dont see currently how :| )

+1 for monospace fonts

akhayrutdinov:
@hcoles, @lkwg82, thanks for replying.
I've updated the PR. ```css``` is now loaded in the constructor, where I also changed two sets creating routines to be slightly more effective.

Also I removed ```font-size``` property, so now the size is inherited from parent container.

hcoles:
@akhayrutdinov Thanks - I've been away without net access for the last week. I'll take a look soon.

Issue #146, opened 2014-10-28
include mutated code in report


UrsMetz:
Currently the mutations are only described in textual form which makes it sometimes hard to find out which kind of mutation of the code is not covered by the tests. I think including the mutated code (like [mutant](https://github.com/mbj/mutant) does in the report in the terminal) for survived mutations in the HTML reports created by PIT could help with that. This would solve issue #145.

hcoles:
I've finally had chance to take a look at this using procyon as @UrsMetz suggested.

Worryingly procyon doesn't look have been updated in a while.

http://www.benf.org/other/cfr/

Looks to be under active development, but is not easy to consume as it is closed source and not published to maven central.

I've put together a proof of concept in the decompilation branch

https://github.com/hcoles/pitest/tree/decompilation

This replaces the mutant description with a diff against the decompiled source.

There are a few issues

1. Currently broken for Java 6 (would need to either drop java 6 support or move into a plugin)
2. Adds a significant overhead. Decompilation will definitely need to be optional.
3. Formatting is terrible.

Not sure the best way to go on point 3. I suspect the best solution would be change the html report so that the decompilation info appears in a dedicated section rather than replacing the existing description.

If we do stick with replacing the description it probably needs some markup added to format it. Doing this is complicated by some tech debt - markdown in the descriptions is currently automatically escaped.

If anyone with better html skills than mine would like to play with this they are welcome.

PR #218, opened 2015-07-18
Naked receiver mutator

UrsMetz:
Mutator that replaces a method call with the object on which the method
is called in case the return type of the method matches the type of the
object.

It is named after @mbj mutation operator in [mutant](https://github.com/mbj/mutant) performing the same mutation.

### DIFF
--- a/pitest/src/main/java/org/pitest/mutationtest/engine/gregor/config/Mutator.java
+++ b/pitest/src/main/java/org/pitest/mutationtest/engine/gregor/config/Mutator.java
@@ -38,6 +38,7 @@ import org.pitest.mutationtest.engine.gregor.mutators.IncrementsMutator;
 import org.pitest.mutationtest.engine.gregor.mutators.InlineConstantMutator;
 import org.pitest.mutationtest.engine.gregor.mutators.InvertNegsMutator;
 import org.pitest.mutationtest.engine.gregor.mutators.MathMutator;
+import org.pitest.mutationtest.engine.gregor.mutators.NakedReceiverMutator;
 import org.pitest.mutationtest.engine.gregor.mutators.NegateConditionalsMutator;
 import org.pitest.mutationtest.engine.gregor.mutators.NonVoidMethodCallMutator;
 import org.pitest.mutationtest.engine.gregor.mutators.RemoveConditionalMutator;
@@ -152,6 +153,11 @@ public final class Mutator {
      */
     add("EXPERIMENTAL_ARGUMENT_PROPAGATION",
         ArgumentPropagationMutator.ARGUMENT_PROPAGATION_MUTATOR);
+
+    /**
+     * Experimental mutator that replaces method call with this
+     */
+    add("NAKED_RECEIVER", NakedReceiverMutator.NAKED_RECEIVER);
 
     addGroup("REMOVE_SWITCH", RemoveSwitchMutator.makeMutators());
     addGroup("DEFAULTS", defaults());
--- a/pitest/src/main/java/org/pitest/mutationtest/engine/gregor/mutators/NakedReceiverMutator.java
+++ b/pitest/src/main/java/org/pitest/mutationtest/engine/gregor/mutators/NakedReceiverMutator.java
@@ -23,6 +23,9 @@ import org.pitest.mutationtest.engine.gregor.MethodInfo;
 import org.pitest.mutationtest.engine.gregor.MethodMutatorFactory;
 import org.pitest.mutationtest.engine.gregor.MutationContext;
 
+import static org.objectweb.asm.Opcodes.POP;
+import static org.objectweb.asm.Opcodes.POP2;
+
 /**
  * Mutator for non-void methods whos return type matches
  * the receiver's type that replaces the method call with the receiver.
@@ -78,7 +81,7 @@ public enum NakedReceiverMutator implements MethodMutatorFactory {
             .registerMutation(this.factory,
                 "replaced call to " + owner + "::" + name + " with receiver");
         if (context.shouldMutate(newId)) {
-          // remove call to method by swallowing the visitMethodInsn call
+          popMethodArgumentsFromStack(desc);
           return;
         }
         this.mv.visitMethodInsn(opcode, owner, name, desc, itf);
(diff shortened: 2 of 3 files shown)


mbj:
I can confirm this mutation is called that way, and it helps a lot in catching specification holes in method changes.

UrsMetz:
Thanks @mbj for confirming the name (that was one reason for mentioning you ;-)).

mbj:
> Thanks @mbj for confirming the name (that was one reason for mentioning you ;-)).

BTW mutant does not give the mutation operators explicit names internally. I simply walk the AST and recursively, per node (while considering the context) emitting mutations.

https://github.com/mbj/mutant/blob/master/lib/mutant/mutator/node/send.rb#L108-L115

I consider to give mutation operators a name, but as ruby has ~100 nodes, and I've 1-4 dedicate mutation operators for that given node, this might be overwhelming. What might work is grouping the operators per class.

mbj:
@UrsMetz Sorry when I ask this again, are you also a mutant user? Your PRs to pitest show deeper mutant exposure.

hcoles:
@mbj The naming is important for pitest as it can't use diffs to explain what the mutation is - all the user has to go on is the mutant description.

UrsMetz:
@mbj I use mutant in my private Ruby pet projects and when I trying to convince people that mutation testing is a cool tool ;-) because I like the simplicity and the power/large number of mutation operators that mutant offers.

But I don't know whether I have deeper mutant exposure (because I only use mutant in private projects as my "job language" is Java) I'm just comparing the tools and looking for having them on the same level :-).

mbj:
> @mbj The naming is important for pitest as it can't use diffs to explain what the mutation is - all the user has to go on is the mutant description.

Good point, and it makes lot of sense for pitest. I consider to do this for documentation proposes, because many people raise issues for mutant saying "mutation 'foo' is not valid", if I'd have a set of descriptive names, I could refer users to the documentation, instead of having to repeat my self over and over agan.

> But I don't know whether I have deeper mutant exposure (because I only use mutant in private projects as my "job language" is Java) I'm just comparing the tools and looking for having them on the same level :-).

I do not think mutant and PItest are directly comparable. Because of the different env constraints. Both tools have their weakness and strenth, they currently learn a lot from each other, thats great.

UrsMetz:
@hcoles There's #146 to fix this ;-). I looked into providing a pull for this but I think some refactoring would be needed to get the mutated byte code information into the reporter output.

hcoles:
@UrsMetz Thanks for the PR. I like the operator. Unfortunately it will suffer from type erasure problems when used with generic types - however I think the problem may be less severe than for other mutators. The "full" type should be inferable for a some subset of cases (e.g when the parameterised type is held as a member). It should also be straightforward to simply exclude generic types from this mutation, which might be the most reasonable behaviour.

Could you add some tests cases to describe the current behaviour when used with generic types?

UrsMetz:
@hcoles I also thought about the "generics problem" but then forget because I was too surprised/happy that a (maybe naive) implementation was quite easy :-).
I will write some tests for behaviour when generics come into play. But they might not be ready before next weekend.

UrsMetz:
> I do not think mutant and PItest are directly comparable. Because of the different env constraints. Both tools have their weakness and strenth, they currently learn a lot from each other, thats great.

@mbj From a user perspective they should be or at least as a user I want them to have the same/comparable features.

mbj:
I'm curious, where do I find information about the generics problem?

lkwg82:
After merged, we need an update of the documentation. (I'd like to do this).

Another idea:

What about to write a blog entry on this topic to tell a story when mutation testing makes sense. I think with this new mutator you have some examples in mind.

I'd like to raise some publicity on this topic and maybe attract some new committer?!

hcoles:
@mbj A starting point for reading on type erasure -> http://stackoverflow.com/questions/339699/java-generics-type-erasure-when-and-what-happens

hcoles:
@lkwg82 Thanks! Documentation often gets forgotten.

mbj:
@hcoles thx.

mbj:
@hcoles That type erasure. I thought its a term special to MA, I now see you simply might lack the context to do the naked receiver emit.

UrsMetz:
> The "full" type should be inferable for a some subset of cases (e.g when the parameterised type is held as a member). It should also be straightforward to simply exclude generic types from this mutation, which might be the most reasonable behaviour.

@hcoles I added one test for the case that the receiver is generically typed (inspired by your test case for the argument propagation mutator).
I didn't change the implementation: The mutator is stilling mutating code where the method call returns a different generically typed version of the same "wrapper" type. I tried to come up with some means to infer the generic type and to avoid mutations in such a case but couldn't find one. It seems that ASM's [SignatureVisitor](http://asm.ow2.org/asm50/javadoc/user/org/objectweb/asm/signature/SignatureVisitor.html) provides means to achieve this but I couldn't find a way to provide it with the necessary informations from what is passed to the mutator's `MethodVisitor`.

UrsMetz:
@hcoles I rebased my branch onto the current master and resolved the conflicts.
Anything I can do to get this pull merged (e.g. more tests for the cases when generics kick in)?

hcoles:
@UrsMetz I took a quick look at the code last night.

There's a bug that will be revealed when the mutator is used in classes with multiple possible mutation points. At the moment the mutator will always mutate the class at every possible point. The way pitest works the mutator gets called twice - once in a scan phase, the again passing in the identifier of the mutation it wished to create.

The bytecode should only be mutated when the mutator is passed an id that matches the current mutation point.

UrsMetz:
@hcoles thanks for reviewing. I pushed a fix just now.
I think when I started the mutator I had the `context.shouldMutate(id)` in it but it got lost while I was working with the code (maybe because I didn't fully understand what the line did).
I checked oher mutator and their tests but couldn't find a test whether the check for the current mutation point is in it. Is there any way to test this in a unit test?

hcoles:
Now merged in. Sorry about the short delay.

mbj:
@hcoles this is OSS, I beat you with sum of delays on my project, no need to feel sorry

PR #219, opened 2015-07-21
performance improvements when annotating coverage

tobiasbaum:
I encountered a performance issue when testing PIT at our projects, which did not occur on some small sample projects I tested. After mutation testing was finished, PIT/PITclipse still needed hours to present the results. With profiling, I could trace it back to methods called during the creation of annotated source files (MutationHtmlReportListener.createAnnotatedSourceCodeLines()). Determining the classes contained in a file was very inefficient for large classpaths because some lists were iterated repeatedly. I therefore replaced this structure with a map for the lookups (and made regex matching a little more efficient). This brought runtime down to seconds on my machine.

I would have preferred filling the map in the constructor of CoverageData, but I had to make it lazy because otherwise the mocking that is used in the unit tests for CoverageData did not work anymore.

### DIFF
--- a/pitest/src/main/java/org/pitest/coverage/CoverageData.java
+++ b/pitest/src/main/java/org/pitest/coverage/CoverageData.java
@@ -56,9 +56,13 @@ public class CoverageData implements CoverageDatabase {
 
   private boolean                                             hasFailedTest = false;
 
+  private Option<Map<String, Collection<ClassInfo>>>          classesForFile;
+
   public CoverageData(final CodeSource code, final LineMap lm) {
     this.code = code;
     this.lm = lm;
+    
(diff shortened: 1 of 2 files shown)


lkwg82:
1) I'd like to ask you to post 'cloc' (`apt-get install cloc`) result to get a feeling of your project size. Please add the number of tests.

With these number it may be possible to add an open source project with the same characteristics as test project.

2) What are you times with maven pitest run before/after change?

thx

hcoles:
@tobiasbaum Thanks for this - not the nicest bit of code to be dealing with.

I think it should be straight forward to make the tests pass with a non lazy intialisation of the cache - just move the construction of the testee to a line after the CodeSource mock has been configured.

PR #223, opened 2015-08-14
fixing execution not covered by lifecycle errors

jasonmfehr:
When pulling the pit project into STS, I have several errors for the checkstyle, compiler, and maven plugin plugins where their execution is not covered by a lifecycle.  I fixed the checkstyle related errors using Quick Fix to install the correct m2e connector which then created a .checkstyle file (hence the change to .gitignore).  I could not find a connector for the maven-plugin-plugin, so I ignored its goals.  I have not tackled the errors with the maven-compiler-plugin yet.  I may do so later if I have time.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ target/
 .idea
 *.iml
 dependency-reduced-pom.xml
+.checkstyle
+
--- a/pitest-maven/pom.xml
+++ b/pitest-maven/pom.xml
@@ -58,6 +58,36 @@
 				</executions>
 			</plugin>
 		</plugins>
+        
+        <pluginManagement>
+            <plugins>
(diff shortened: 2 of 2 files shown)


hcoles:
@jasonmfehr My preferred approach is not to pollute the pom with eclipse concerns, but to instead use the (until recently experimental) option of marking the plugins as ignored ion the eclipse preference file.

This has the downside that eveyone needs to do this each time they import the project, but it avoids cluttering the pom.

Unfortunately Eclipse/maven integration is pretty messed up.

jasonmfehr:
I understand not wanting to clutter up the poms.  I will close this pull request.  Will you please consider making the update to add ".checkstyle" to the gitignore file?

PR #240, opened 2015-12-02
using regex in globs

sebi-hgdata:
Hey,
   We have some patterns that don't work well with the current implementation of globs. For example, we have Spock specs in packages like a.b.c.*Spec, a.b.c.integration.*Spec, a.b.c.integration.ui.*Spec, etc...  now the issue is that we want to run specs from the first 2 packages , but not from the last one and using a pattern like 'a.b.c.*Spec' or a.b.c.integration.*Spec (this is a subset of the first one with the current implementation) includes also the a.b.c.integration.ui.*Spec. 
  In this PR we added a simple convention that if the glob starts with ~ (this is the regex operator in Groovy) we handle it as a Pattern regex. 
 Let me know what you think


hcoles:
Thanks - nice pragmatic way off adding regex support which I think a few people have requested.

sebi-hgdata:
Super

PR #245, opened 2016-01-11
Do not split child JVM arguments using commas as delimiters. Without …

cinquin:
…this change, using an Ant task I could not see a way to pass an argument such as -XX:CompileCommand=exclude,gnu.trove.impl.hash.TObjectHash::insertKey (this would presumably require a way of escaping the commas). I cannot think of circumstances under which it would be useful for the arguments to be split by comma, but perhaps I'm missing something. In the case of a command line invocation, it seems that it should be OK for multiple arguments to be separated by spaces as long as the whole argument string is between quotes (not tested).

### DIFF
--- a/pitest-command-line/src/main/java/org/pitest/mutationtest/commandline/OptionsParser.java
+++ b/pitest-command-line/src/main/java/org/pitest/mutationtest/commandline/OptionsParser.java
@@ -169,8 +169,8 @@ public class OptionsParser {
         .describedAs("comma separated list of mutation operators");
 
     this.jvmArgs = parserAccepts(CHILD_JVM).withRequiredArg()
-        .withValuesSeparatedBy(',')
-        .describedAs("comma separated list of child JVM args");
+        .describedAs(
+                "Child JVM args provided as a single string that can include spaces");
 
     this.mutateStatics = parserAccepts(MUTATE_STATIC_INITIALIZERS)
         .withOptionalArg()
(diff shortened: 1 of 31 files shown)


hcoles:
This seems reasonable, but I'd like to hear from @szpak as it is likely to impact the gradle plugin.

szpak:
At the Gradle side the child JVM arguments are configured as a formal list of string (`List<String>`), so it can be converted to any format that is needed. Nevertheless most of parameters in PIT command line syntax assume coma separation. It could be (potentially) misleading for the users. Luckily that argument is (rather) rarely used, so hopefully the impact will not be much.
I would like to confirm only that it will be ok for the PIT parser to pass two jvmArgs arguments (or just one) as a string in quotes:
```
--jvmArgs "-XX:CompileCommand=exclude,gnu.TObjectHash::insertKey -Xmx1024m"
```

cinquin:
> I would like to confirm only that it will be ok for the PIT parser to pass two jvmArgs arguments (or just one) as a string in quotes:
> 
> --jvmArgs "-XX:CompileCommand=exclude,gnu.TObjectHash::insertKey -Xmx1024m"

Yes that works fine for me (from an Ant task at least). Although I gave an example with just one argument to focus on the comma problem, in practice I have been using multiple arguments.

cinquin:
Actually it turns out that, although the multiple arguments were passed to the JVM, they were not parsed correctly by the JVM because the whole string was seen a single argument. I did not catch this straight away because a single argument such as `"-XX:CompileCommand=exclude,gnu.TObjectHash::insertKey whatever junk"` (note the quotes) causes the JVM to display an error message but not to abort. It looks like the error message from the Minion was swallowed up and did not make it to the console.
I have updated my pull request to work around this problem by splitting the arguments by space. Note that the tradeoff here is that this probably makes it impossible to pass space-containing arguments to the JVM. I guess there probably is a more elegant solution to this problem, but at least this works and I don't know off-hand of a use-case in this context for space-containing JVM arguments.

maxgabut:
Closing this as it has been inactive for a while now.

Issue #263, opened 2016-05-01
Bug: tests with @RunWith do not end up in the line coverage


theangrydev:
This means they are never used for mutation testing.

This happens when using @RunWith(TableRunner.class) from https://github.com/bodar/yatspec

theangrydev:
With verbose turned on, I can see that the correct runner seems to be invoked, because I can see the `FINE: Gathering coverage for test...` message with name set to the correct name of the test when the `com.googlecode.yatspec.junit.TableRunner` is used

theangrydev:
And if I rewrite the test to still use `TableRuner` but not use a parametrised `@Table` test then the coverage is correctly picked up.

hcoles:
Thanks for the PR.

As you've probably noticed the current logic is the wrong way round - there should really be a whitelist of runners that should be split rather than a blacklist of ones that shouldn't.

If I get chance I'll implement this change, otherwise I'll merge this in before the next release.

hcoles:
Released in 1.10

PR #264, opened 2016-05-02
Quick fix for compatibility with yatspec runners

theangrydev:
This is a quick fix for #263 

As a long term solution, I think the list of runners in runnerCannotBeSplit should be part of the configuration


theangrydev:
I would appreciate a release for this if you can please

Issue #170, opened 2015-02-10
False error: Mutation testing requires a green suite


Kidlike:
I am running pitest against a green suite.
Though specific test classes fail when run with pitest.

The failing test classes have the following annotations:

```
@RunWith(SpringJUnit4ClassRunner.class)
@ContextConfiguration(locations = "/domain-test-context.xml")
```

This setup is being used so that we test code against a real DB.

If I am not mistaken these annotations is the reason these classes fail when being run by pitest.

Is this solveable ?

If not then is it possible to somehow configure pitest to ignore such tests ?

Kidlike:
When I say ignore such tests, I mean have a section in the pitest configuration like the `avoidCallsTo`. Something like `ignoreTestsThatUse`.

hcoles:
Pitest has no problem with spring annotations that I'm aware of, but if these tests are hitting a real db then they are not suitable for mutation testing as they unlikely to be repeatable.

Tests can be excluded using the excludeClasses option.

There isn't currently a built in way to exclude tests by annotation - although pitest does have partial support for junit categories that could be used for this purpose.

Kidlike:
jUnit categories seem interesting. Thanks !

***About the actual fail :***

I should have posted this from the beginning.
It throws this for several test cases:
```
5:52:59 PM PIT >> FINE : SLAVE : FAIL Description [testClass=com.company.propagation.domain.DataQueueSpringTest, name=testSerializationAndDeserialization(com.company.propagation.domain.DataQueueSpringTest)] -> java.lang.IllegalStateException: Failed to load App
```

And yes, the tests that are failing have these annotations, and are hitting a real db.
About the repetition, the db is cleared in the `@After` section. Shouldn't that be sufficient ?

KyleRogers:
That depends on several possibilities:
- if you run pitest with more than 1 thread those tests might still interfere with each other
- if the cleanup code is part of the production code it might get mutated and clean up might fail
- it might not be the database but the file system or another shared resource which is not well isolated


Nevertheless the line you've postet ends with "Failed to load App[lication Context]" which is more a sign that the integration test(s) couldn't start at all. Maybe the database is not ready yet, maybe there are class path issues or maybe there is another problem.

In my opinion mutating integration tests - and if you hit a database those tests are integration tests not unit tests - is dangerous because those tests are designed to interact with the environment (file system, database, ...) and you cannot tell what a mutant might do to those resources. Maybe one of the mutants erases the whole file system....

[...]

PR #283, opened 2016-07-06
Extension for Issue #170

galovics:
Extension to support the test class exclusions based on the test class RunWith annotation.

Now the configuration is extended with the excludedRunners property which is a collection of runners.

### DIFF
--- a/pitest/src/main/java/org/pitest/classpath/CodeSource.java
+++ b/pitest/src/main/java/org/pitest/classpath/CodeSource.java
@@ -9,6 +9,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 
+import org.junit.runner.RunWith;
 import org.pitest.classinfo.ClassInfo;
 import org.pitest.classinfo.ClassInfoSource;
 import org.pitest.classinfo.ClassName;
@@ -18,6 +19,7 @@ import org.pitest.classinfo.TestToClassMapper;
 import org.pitest.functional.F;
 import org.pitest.functional.FCollection;
 import org.pitest.functional.Option;
+import org.pitest.functional.predicate.Predicate;
 import org.pitest.testapi.TestClassIdentifier;
 
 /**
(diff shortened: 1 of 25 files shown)


pitest/src/main/java/org/pitest/classpath/CodeSource.java
```diff
@@ -57,7 +63,9 @@ public CodeSource(final ProjectClassPaths classPath,
   public List<ClassInfo> getTests() {
     return flatMap(this.classPath.test(), nameToClassInfo()).filter(
         and(isWithinATestClass(), isIncludedClass(),
-            not(ClassInfo.matchIfAbstract())));
+            not(ClassInfo.matchIfAbstract()),
+            not(ClassInfo.matchIfAnnotationPresentWithValueString(RunWith.class,
```
KyleRogers: I don't think it's a good idea to check this in CodeSource. I think we should move that to JUnitTestClassIdentifier.isIncludedClass()

This is JUnit specific behaviour and shouldn't be executed in TestNG setups. I'm surprised that it seems to run inside a TestNG setup. But Annotations are usually treated differently by the class loader.

KyleRogers:
I'm missing an integration test or acceptance test that demonstrates that the new feature is working as expected. I'm sorry I had not the time to find a suitable Test class to point you to.

pitest/src/main/java/org/pitest/classpath/CodeSource.java
galovics: I'll check it out, thanks!

galovics:
Because I haven't added any tests. First, I wanted someone to review the code that this is the direction or not. After I fixed the mentioned issue, I'll add some. :-)

galovics:
The logic went to the JUnitTestClassIdentifier as mentioned plus added tests to verify the functionality.

galovics:
any update here? :-)

pitest/src/main/java/org/pitest/junit/JUnitTestClassIdentifier.java
```diff
@@ -39,7 +42,16 @@ public boolean isATestClass(final ClassInfo a) {
 
   @Override
   public boolean isIncluded(ClassInfo a) {
-    return isIncludedCategory(a) && !isExcludedCategory(a);
+    return isIncludedCategory(a) && !isExcludedCategory(a) && isNotRanWithExcludedRunner(a);
+  }
+
+  private boolean isNotRanWithExcludedRunner(ClassInfo a) {
+    String runWith = getRunWithAnnotationValue(a);
+    return this.excludedRunners.isEmpty() || !this.excludedRunners.contains(runWith);
```
hcoles: Why is the additional call to isEmpty required here?

galovics: To be honest, I don't really know why I put that there. Probably it was part of my previous implementation. :-)

Gonna remove it.

hcoles:
I had a quick scan through the diffs last night and it seems reasonable enough, but what is the use case this is supporting? What's the rationale behind filtering out tests by runner?

galovics:
Please check [this](https://github.com/hcoles/pitest/issues/170#issuecomment-230692754) thread from this comment

hcoles:
I don't think we need to worry too much about changing the api - we'd up the version to 1.2.0 but in reality I don't think anyone will really care.

My main concern about this PR is I'm not convinced the functionality will be useful for most people. My preference would be implement this via a plugin rather than as a feature of the main codebase, however I don't think a suitable extension point currently exists.

galovics:
This can be implemented as a plugin of course. However, if there is no such extension point yet, we can include this in the main codebase and when the plugin support is there, we can migrate it.

galovics:
Thank you for merging it. May I ask you when do you plan the next release? :-)

PR #287, opened 2016-07-29
Look for JUnit annotations up in class hierarchy too

Kwaq:
It would be useful if @Category and @RunWith annotations placed on test superclasses could be considered too.


hcoles:
@Kwaq Does this match how categories behave with surefire?

Kwaq:
Yes. Since JUnit 4.12 @Category has @Inherited added (https://github.com/junit-team/junit4/issues/398).

hcoles:
I think, to match normal junit behaviour, we would need to check that the version of `@Category` on the classpath was annotated with `@Inherited` before navigating up the hierarchy.

hcoles:
Thanks for the PR - sorry it took so long to merge

Issue #285, opened 2016-07-26
Problem with String in switch statement


Kwaq:
RemoveConditionalMutator_EQUAL_IF marks switch on String statement as survived.

Method

``` java
public String switchString(final String input) {
        final String result;

        switch (input) {
            case "a":
                result = "A";
                break;
            case "b":
                result = "B";
                break;
            default:
                throw new IllegalArgumentException("Unsupported input");
        }

        return result;
    }
```

and tests

``` java
 @Test
    public void stringSwitchShouldReturnA() throws Exception {
        // given
        final String input = "a";

        // when
        final String result = switchExample.switchString(input);

        // then
        assertThat(result).isEqualTo("A");
    }

    @Test
    public void stringSwitchShouldReturnB() throws Exception {
        // given
        final String input = "b";

        // when
        final String result = switchExample.switchString(input);

        // then
        assertThat(result).isEqualTo("B");
    }

    @Test(expected = IllegalArgumentException.class)
    public void stringSwitchShouldThrowIllegalArgumentException() throws Exception {
        // given
        final String input = "c";

        // when
        switchExample.switchString(input);

        // then
        // exception
    }
```

Output report:

```
Mutations

25      
1. removed conditional - replaced equality check with false → KILLED
2. removed conditional - replaced equality check with false → KILLED
3. removed conditional - replaced equality check with true → SURVIVED
4. removed conditional - replaced equality check with true → SURVIVED
```

Code for analogous integer switch is not mutated.

 Code for this case is available at [my repo](https://github.com/Kwaq/pitest-issue)

hcoles:
Thanks for the high quality bug report - unfortunately this is going to be a bit tricky to fix.

The issue is that when switching on strings the compiler generates code something like

```java
int hash = someString.hashCode();
switch (hash) {
  case 97 :
    if (someString.equals("a") {
     return "A"; 
   }

  case 98 :
    if (someString.equals("b") {
     return "B"; 
   }
}
```

To kill the mutant in the equality check you'd have to feed it an unequal string with the same hash code as "a".

The "correct" behaviour here is probably to filter out any conditional mutations that relate to a string switch statement.

I'm not sure if this is possible (or at least not without sometimes filtering out valid mutations). Some experimentation would be required - a hack that's been used in other bits of pitest is to only trigger filters where the line number is 0 or (as might work in this case) where large amounts of code share the same line number.

A mutation filter 

[...]

PR #288, opened 2016-08-02
Implements #285

Kwaq:
RemoveConditionalMutator_EQUAL_IF creates mutations for switch on string value and this mutations cannot be killed. This is fixed by implementing filter that removes this particular mutation in case when
it was done on switch instruction.

### DIFF
--- a/pitest-java8-verification/src/test/java/com/example/java7/HasIfOnAStringEqualityInsideSwitchTestee.java
+++ b/pitest-java8-verification/src/test/java/com/example/java7/HasIfOnAStringEqualityInsideSwitchTestee.java
@@ -1,0 +1,28 @@
+package com.example.java7;
+
+public class HasIfOnAStringEqualityInsideSwitchTestee {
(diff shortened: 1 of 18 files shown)


hcoles:
Thanks, I've taken a quick scan and it looks good. 

I'm a little tied up with some things at the moment so probably won't be able to take a proper look for 2 weeks (sorry), but I've a couple of quick questions based on the scan.

How does it deal with with programmer written if statements within switches? Would these also be filtered?

I think there are also a few logic statements of the form

```
if (x) {
 if (y) {
  return true;
 }
  return false;
} else {
  return false;
}
```

That could be condensed down to

`return x && y;`

Kwaq:
I'll try and test your cases.

hcoles:
Ok, after not even a full year of waiting I've merged this in.

Having looked into the details of I changes the approach to not use a filter but instead do something a little lower level using the asm sax api. Although the final approach is different it wouldn't have happened without the work put into the PR, and the tests are still used to validate it - thanks.

PR #291, opened 2016-08-05
Handle empty excludes

### LINKED ISSUE #290: pitest-maven fails since 1.1.8 when Surefire exclude is empty string
If Surefire is configured with an exclude that is the empty string (which is something that happens in our poms, since we set it to exclude a configurable pattern, and the property that holds the pattern is empty when nothing should be excluded), then the convertExcludes fails due to Glob trying to read the first character of an empty string, yielding a java.lang.StringIndexOutOfBoundsException.
This problem first appeared in 1.1.8, and is still present in master as of db2e0dfbb.

magnusr:
Fixes #290

### DIFF
--- a/pitest/src/main/java/org/pitest/util/Glob.java
+++ b/pitest/src/main/java/org/pitest/util/Glob.java
@@ -26,7 +26,7 @@ public class Glob implements Predicate<String> {
   private final Pattern regex;
 
   public Glob(final String glob) {
-    if (glob.charAt(0) == '~') {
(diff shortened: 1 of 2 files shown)


hcoles:
@magnusr Thanks - if you can switch that length > 0 to a !isEmpty() I'll merge it in.

magnusr:
Switched to `.startsWith` instead.

hcoles:
That's a much better idea.

PR #296, opened 2016-09-02
add kotlin verification and filter kotlin intrinsics

christophsturm:
My kotlin test duplicates a lot of code because i did not want to refactor your code. I guess createEngineAndRun and the other 2 methods could be made static and optionally moved. 

Also I'm not sure if AvoidAssertsMethodAdapter is the right place for this check, or if it should be a separate adapter. But for now it works and the kotlin code is basically an assert.

### DIFF
--- a/pitest-kotlin-verification/pom.xml
+++ b/pitest-kotlin-verification/pom.xml
@@ -1,0 +1,71 @@
+<?xml version="1.0"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <artifactId>pitest-parent</artifactId>
+        <groupId>org.pitest</groupId>
+        <version>1.1.11-SNAPSHOT</version>
+    </parent>
+    <artifactId>pitest-kotlin-verification</artifactId>
+    <packaging>jar</packaging>
+    <name>pitest-kotlin-verification</name>
+    <description>tests for pit kotlin integration</description>
+    <url>http://pitest.org</url>
+    <properties>
+        <kotlin.version>1.0.3</kotlin.version>
+    </properties>
+    <build>
+        <testSourceDirectory>${project.basedir}/src/test/kotlin</testSourceDirectory>
+        <plugins>
+            <plugin>
+                <artifactId>kotlin-maven-plugin</artifactId>
+                <groupId>org.jetbrains.kotlin</groupId>
+                <version>${kotlin.version}</version>
+
+                <executions>
+                    <execution>
+                        <id>compile</id>
+                        <goals>
+                            <goal>compile</goal>
+                        </goals>
+                    </execution>
+
+                    <execution>
+                        <id>test-compile</id>
+                        <goals>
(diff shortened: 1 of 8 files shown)


hcoles:
Thanks for this.

As much as possible I'd like to keep the code to support kotlin separate and easily identifiable, so I'd prefer it moved into a separate method adapter. 

It'd also be good to add some unit level tests to cover the new behaviour. The ones in the kotlin verification project are valuable but will give a slow feedback cycle when things change.

christophsturm:
makes sense. I didn't write unit tests because i wanted to make sure it's in the right place. How can I add another visitor that filters out the kotlin parts?

also are you ok with making the test methods static an reusable?

hcoles:
The filters are added in a very straight forward manner - a hard coded chain of them is created in the MutatingClassVisitor

https://github.com/hcoles/pitest/blob/8f654ad3849d5dc5b7cbe59dddea788d19ded317/pitest/src/main/java/org/pitest/mutationtest/engine/gregor/MutatingClassVisitor.java#L123

Yes, make the methods static if that makes sense.

christophsturm:
I moved the code to a new MethodVisitor(mostly copy pasted from the AvoidAssertsMethodVisitor)
.
 I noticed that there's already a KotlinFilter, but my kotlin verification was for some reason not running with filters, so I decided against putting the code there. 

Is there a rule what should go into a filter and what should go into a method visitor?

hcoles:
Generally if you can sensibly implement it as a filter then that's where it should go, so adding this to the existing kotlin filter is probably the thing to do.

christophsturm:
ok how do i need to check my integration test to make it use filters?

hcoles:
Sorry for the very long delay - finally got a chance to look at this properly.

The reason the integration test doesn't pick up mutation filters is that it goes in at too low a level. The TestMutationTesting class duplicates a subset of what happens if you go in through the normal EntryPoint class. That subset does not include discovering and loading filters.

The current solution for writing full end to end integration tests is to write maven verification. Something that went in at a slightly lower level via the EntryPoint class would be a good idea, but it doesn't currently exist.

There looks to be a bug in the current code. The AvoidKotlinMethodAdapter will disable mutations but not enable them again.

So I think adding this to the existing kotlin filter is the way to go. It's a bit hacky as all we can do is check for "kotlin/jvm/internal/Intrinsics" in the mutation description - but the existing kotlin support is no less hacky.

christophsturm:
my motivation for this pr is to lay the groundwork for kotlin support that is not hacky. I agree that we should use the filter support if thats enough for the current filtering, but at some point we will probably need a method adapter anyway. 

I really don't want to mess with maven + kotlin, but I also want this tested, so why don't we keep it in the adapter for now (and maybe move the other kotlin support to the adapter too). Or you could help me write the testing support via the EntryPoint class.

hcoles:
Thanks for your work on this, I'm closing the PR as there is now the start of a filter for kotlin in the pitest org.

https://github.com/pitest/pitest-kotlin

This still needs some-one to work on it, but there are some quicks wins that could be had for very little effort filtering out intrinsics calls.

christophsturm:
oh thats great. i have been using pitest with kotlin pretty succesful just with     `avoidCallsTo = ["kotlin.jvm.internal"]` but will try out the plugin

hcoles:
I think the plugin will currently be less effective as it is unfinished work, but could be easy improved (eg be adding a filer for kotlin.jvm.internal).

christophsturm:
ok but can i use the plugin and the avoid calls setting together?

hcoles:
Can't see why not.

PR #304, opened 2016-10-07
Merging the updates from the original PiTest

shishirk2016:
```
1. Write mutated class files to the disk.
2. Ignore non-green test cases and continue with the mutation for rest of the suite.
```


hcoles:
Closing as appears to have been created in error.

PR #305, opened 2016-10-08
Make the stack trace visible in the logs when a test fails without mutation

theangrydev:
This will help to figure out why tests are failing without mutation


marcinMarcin:
It would be helpful 👍

davidburstromspotify:
Indeed

PR #327, opened 2017-01-30
Use Implementation-Vendor-Id instead of Implementation-Vendor

oscarlvp:
Changed the code of DependencyFilter so plugins are matched against Implementation-Vendor-Id instead of Implementation-Vendor. This way the plugin discovering is compliant with the default behavior in maven that sets groupId to Implementation-Vendor-Id in the META-INF/MANIFEST.MF


oscarlvp:
Thanks :)
I been working also in trying to make the tool work as an aggregator in
multi-module maven projects. So far, what I have done is to gather and
combine the classpath of all subprojects and invoke the command line tool.
I think this could be implemented as a goal.
​

Issue #372, opened 2017-07-13
Introduce rounding strategy, or mechanism to specify decimal positions for rounding


paul-hammant:
I'd like to see percentages rounded to say two or three decimal positions. #140 is around this topic too.

It'd be nice to control that.

If you're interested in why?  I've 1000 tests of dubious quality. We're going to tell the dev team to 1) improve those existing ones over time, but also 2) make sure that new ones are perfect from the outset.  We're going to put the figures in a time series database and chart it. That'll have zoom in zoom out and pan features.  We want to speak to development if the percentages even goes down 0.01% because that'll be a clue perhaps that development is transgressing on (2).

PR #373, opened 2017-07-15
Dec Positions for over all percentage

paul-hammant:
relates to #372 - stop rounding and have three decimal positions

### DIFF
--- a/pitest-entry/src/main/java/org/pitest/mutationtest/statistics/MutationStatistics.java
+++ b/pitest-entry/src/main/java/org/pitest/mutationtest/statistics/MutationStatistics.java
@@ -49,23 +49,23 @@ public final class MutationStatistics {
     return getTotalMutations() - getTotalDetectedMutations();
   }
 
-  public long getPercentageDetected() {
+  public float getPercentageDetected() {
     if (getTotalMutations() == 0) {
-      return 100;
+      return 100f;
     }
 
     if (getTotalDetectedMutations() == 0) {
-      return 0;
+      return 0f;
     }
 
-    return Math.round((100f / getTotalMutations())
-        * getTotalDetectedMutations());
+    return (100f / getTotalMutations())
+        * getTotalDetectedMutations();
   }
 
   public void report(final PrintStream out) {
     out.println(">> Generated " + this.getTotalMutations()
         + " mutations Killed " + this.getTotalDetectedMutations() + " ("
-        + this.getPercentageDetected() + "%)");
+        + new DecimalFormat("#.###").format(this.getPercentageDetected()) + "%)");
     out.println(">> Ran " + this.numberOfTestsRun + " tests ("
         + getTestsPerMutation() + " tests per mutation)");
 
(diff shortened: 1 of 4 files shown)


hcoles:
Thanks for the PR. 

I'll take a proper look when I'm not thumb typing on a phone - but rounding has been an emotive issue for people in the past so some thought might be required.

paul-hammant:
Hey, any thoughts about merging this in, Henry?

hcoles:
Sorry - only just looked.

I think this may need a little more work before it can be merged in. The switch to float means that the an int/float comparison is now taking place when enforcing the mutation threshold.

I haven't quite got my head round what the result of this will be, but to preserve the existing behaviour I think the float will need to be rounded as per the original code when applying the threshold.

The comparison is currently duplicated in the commandline tool and maven plugin (and presumably also the gradle plugin). It should perhaps be made a responsibility of the MutationStatistics class to remove the duplication.

paul-hammant:
About ten years ago I donated a JSON capability to Joe Walnes for his ground breaking [XStream](http://x-stream.github.io/) technology.  No pull-requests back then. He consumed it, then rewrote all of it in a single Svn commit, while still being generous and crediting me.  

At some point, IBM called me to ask about IP as they were using XStream in something that required them to sleuth through the contributors. A little embarrassed, I cooperated fully and confirmed that I was free to make contributions and not restrained by any IP arrangement with my employer (ThoughtWorks).  

I'm quite used to authors liking the idea but moving to the correct implementation :)

More on Joe and open source: https://paulhammant.com/2014/05/26/joe-walnes-on-successful-open-source/

paul-hammant:
> I think this may need a little more work before it can be merged in.

The way Joe would do it would be to merge the change in, then do a follow up commit to improve the float/int logic. And if we're also doing the always release mantra of Joe, we wouldn't mind if there was a release between the one that went out with **this** PR, and the commit that would improve the float/int logic.  i.e. Releasing and consuming community contributions are the most important things for OSS techs that have not jumped to that "have a life of their own" place.

paul-hammant:
Any further thoughts?

hcoles:
The worry was not that the float/int logic would need improving, but that it would result in broken behaviour. Although pitest doesn't release all that frequently I try to keep master in a shippable state.

I've finally found time to check this branch out and try it out (sorry for the slow response). The float/int comparison does not seem to cause any major problems - the limits are still enforced sensibly.

The increased length of the figures does however cause some rendering issues on the html report. If these can be fixed I'd be happy to merge it in.

paul-hammant:
https://paulhammant.com/2014/05/26/joe-walnes-on-successful-open-source/ --> 

.8 Implore potential collaborators to contribute patches (pull requests) that are implementation and tests
.9 Rewrite/Refactor as you speedily consume patches

^ Me writing in 2014 about open source contribution events in 2006.

I'm not going to do any further work on that pull-request. Over to you.

UrsMetz:
@paul-hammant I think your reaction is a bit harsh. The mode you refer to might be nice for the (occasional) contributors.
But I think there are two main reasons for maintainers to follow a policy that differs from your proposed policy like @hcoles does (apart from the releasable master argument):
* restricted time: reworking contributions can be time consuming
* building a contributor basis: I think some people would be a bit upset or confused when their recently merged changes are turned inside out and would rather have been guided to the solution the maintainer prefers. From the project I follow on GitHub and contributed to I'm used to this kind of policy.

paul-hammant:
I regret doing this work. I thought I had deleted it. I deleted my fork that's for sure. The author should just ignore this and do what he wants to do, or not do what he wants to do.

PR #378, opened 2017-07-21
CSS cleanup for HTML reports

cosminrentea:


hcoles:
Thanks - the html and css are indeed a complete mess.

PR #408, opened 2017-10-02
correct issue with SCM-Goal in multi-module project

sarahBuisson:


hcoles:
Works nicely - thanks.

PR #409, opened 2017-10-02
run only the mutation on the src who are in the origin branch but not in the destination branch

sarahBuisson:
with the param originBranch and destinationBranch, you can run only the mutation test in the modified code. 

By example, It will allow to run the mutation test only on the impacted code of a PR. Moderne build factory can publish this report as comment of the PR ( or at least a link to the repport)

I test-it in my  project,(it work !) but I don't have any idea how create a IT test for this .


hcoles:
Yes, testing this stuff is hard - none of the SCM functionality currently has a meaningful integration test. I think the solution probably involves docker.

Will take a look at the changes when I get chance - thanks for the PR.

sarahBuisson:
I've publish a project who test-it : https://github.com/sarahBuisson/pitest-scm-test
Each branch have a  type of modification with the master.

hcoles:
Thanks - merged in now. Sorry for the delay·

PR #413, opened 2017-10-07
load css from separate file to allow styles to be displayed in jenkins

kiftio:
to fix issues 403 & 256

- Remove styles from html template and append to style.css
- Reference external stylesheet from html template
- Write out style.css when writing index page

The trickier/uglier part... Setting coverage complete width (green bar) without javascript or inline-styles.  I've added a class for each percentage in style.css and then setting the classname in the template.


hcoles:
Thanks for this - sorry for the delay in feeding back.

It all looks good, except there's some regression in the "Mutations" section at the end of the per file report. It seems to have lost the red green colour coding.

Was this a deliberate change?

kiftio:
Ah, there was a case-mismatch between the classname in the css and html files.  It seems to be case-insensitive without the html5 doctype declaration, but case-sensitive with it.  

I've updated the CSS file

hcoles:
Thanks - now merged in.

Issue #422, opened 2017-11-28
Java 9 illegal reflective access to TreeMap.comparator


Pr0methean:
When running on Java 9, I get this warning repeatedly:
```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.pitest.reloc.xstream.core.util.Fields (file:/home/travis/.m2/repository/org/pitest/pitest/1.2.4/pitest-1.2.4.jar) to field java.util.TreeMap.comparator
Wstderr  : ARNING: Please consider reporting this to the maintainers of org.pitest.reloc.xstream.core.util.Fields
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denistderr  : ed in a future release
```

Pr0methean:
Since this class appears to actually be  com.thoughtworks.xstream.core.util.Fields renamed (http://www.docjar.com/html/api/com/thoughtworks/xstream/core/util/Fields.java.html), you may be able to fix this by updating the dependency version of Xstream.

hcoles:
Pitest currently relies on xstream heavily, unfortunately it doesn't seem likely that xstream will be updated to avoid the reflective access warning (which will eventually become an error) any time soon.

The plan is to migrate away from xstream. An (almost) xstreamless version of pitest lives in this branch

https://github.com/hcoles/pitest/tree/no_xstream

It's a pretty major change, and involves dropping the mutate static initializers feature. It also requires making some of the pitest classes serializable, which feels a little dirty.

I've not worked out exactly when this will be merged in. In needs to be soon as keeping a long lived branch going is no fun, but I want to be sure it's the right decision before taking the plunge.

cordisvictor:
@hcoles very late comment, but maybe this could suite the needs described above https://github.com/cordisvictor/easyml-lib

PR #423, opened 2017-11-28
Update Xstream to 1.4.10

Pr0methean:
Fixes https://github.com/hcoles/pitest/issues/422?


Pr0methean:
Update: https://travis-ci.org/hcoles/pitest/jobs/308605637 shows that this does not fix the bug.

PR #433, opened 2017-12-15
Adding an additional report-aggregate mojo

rchargel:
Works similarly to the mojo of the same name from Jacoco.

The idea here is that if you have a multi-module maven project, you can generate individual reports for each of your modules, but then have a "report-aggregation" module which can be used to build a final aggregated report.

The report-aggregation module would have to include the modules it is interested in as dependencies.

As an example, say I have some project called company-functions, that had a few sub-modules:

company-functions
   |____ module-1
   |____ module-2

A third module would be created, maybe "mutation-report-aggregation" , which has module-1 and module-2 as dependencies.

That mutation-report-aggregation module would use the report-aggregate goal for your maven plugin, and an aggregated report would be generated.

The only caveat is that both the Line Coverage report has to have been exported, and an XML formatted formatted version of the mutation report must be present.


szpak:
@rchargel Thanks for looking at the (long awaiting) feature. One high level question from the person maintaining the Gradle plugin for PIT. Does that code have to be Maven dependent? Having it implemented in PIT itself would make it easier to reuse in Gradle/Sbt/Ant/...

rchargel:
I can look at moving it out to one of the other modules after the holiday.

szpak:
@rchargel Great, it would be useful. Nevertheless, it would be good to get an input from @hcoles before that.

rchargel:
@szpak and @hcoles As per the request above, I updated the PR to separate out the aggregation code into a separate module (pitest-aggregator). That module has one important class, ReportAggregator with a builder. The builder is provided with a list of source directories, compiled code directories, and the mutations and linecoverage XML files, and then the aggregator can build an aggregated version of the report.

The "report-aggregate" maven goal now just calls into the code above.  This change would be very useful to my organization as most of our maven builds are separated into 10 or more sub-modules.

Thank you,
Rafael

hcoles:
@rchargel Thanks for the PR - I've not had chance to look at it yet (sorry), but I should finally have some time this weekend.

hcoles:
Finally merged - sorry for the delay

hcoles:
@rchargel Could you create a PR against https://github.com/hcoles/pitest-site that adds documentation for the new mojo?

rchargel:
Sure, I'll try to get to that this week.

sarahBuisson:
@rchargel Hello, I've trouble to use your plugin, have you a sample somewere?

rchargel:
Yes, I owe some documentation. I'll put that together this week.

JustAnotherSoftwareDeveloper:
Can you write this documentation, or direct me to it. I've tried to generate an aggregate report based off the source code, but it isn't working.

rchargel:
I have a PR in

Issue #437, opened 2018-01-10
Filtering executed methods of targeted tests


ftrautsch:
Hi,
for a research project of mine I needed to be able to filter the executed tests of a test class. I did not find any way to accomplish this using the current development version of pitest. Hence, I implemented a new configuration parameter for pitest in which I am able to put in a name (or names) of a test method and pitest only executes these methods for the targeted tests. 

For example, I have two test classes (ClassATest and ClassBTest) and ClassATest has the test methods "foo" and "bar" and ClassBTest has the test methods "bar" and "bar2". If I now set the new configuration parameter to "bar" and targetTests to ClassATest, pitest will only execute ClassATest.bar(). If I also include ClassBTest, it will execute the method bar of ClassATest and ClassBTest. 

I am not sure if this kind of feature is also interesting for other researchers or user of pitest. Therefore, I wanted to create this issue here to see if it makes sense to create a new pull request or not.

hcoles:
I think this would be a useful feature but it might not be as straight forward as it seems. What constitutes a test "method" varies depending on the test frame work, runner etc. Concepts such as BeforeClass and AfterClass might also complicate things.

ftrautsch:
That is true. My idea was to inject the code into the *TestUnitFinder classes. There, I implemented something similar to the Category-based filtering mechanism.

After the list of TestUnits in the "findTestUnits" method is created, I filter them based on their name (see below)

```java
private List<TestUnit> filterUnitsByMethod(List<TestUnit> filteredUnits) {
    if (this.includedMethods.isEmpty()) {
      return filteredUnits;
    }

    List<TestUnit> units = new ArrayList<>();
    for (TestUnit unit: filteredUnits) {
      if (this.includedMethods.contains(unit.getDescription().getName().split("\\(")[0])) {
        LOG.info("Included: " + unit.toString());
        units.add(unit);
      }
    }
    return units;
  }
``` 

As I am not that familiar with all the corner cases that might exist this solution might not work for everything. But, I tried it out with JUnit and some projects and it seems to work fine (at least the results are as expected). What do you think about this approach?

hcoles:
Seems pretty reasonable.

maxgabut:
Closing this issue because it's now fixed. I created an issue to add the documentation.

PR #438, opened 2018-01-11
Added configuration parameter to filter executed methods of target tests 

ftrautsch:
This is a pull request for issue #437 

I tried to be as close as possible to the naming convention and coding style of the project. I also made sure that the tests are running and I created a few new tests for the change.

The goal of this change is to filter the executed tests of a test class.  Hence, I added a new configuration parameter (includedTestMethods) for pitest in which you can put in a name (or names) of a test method and pitest only executes these methods for the targeted tests. The idea is to change the  *TestUnitFinder classes. There, I implemented something similar to the Category-based filtering mechanism.


hcoles:
Thanks - sorry for the delay merging.

PR #445, opened 2018-02-11
Raise minimum supported java version to 8

hcoles:
Drops support for Java 7, converts anonymous inner classes to lambdas and replaces parts of the org.pitest.functional package with java.lang classes.


README.md
```diff
@@ -9,7 +9,7 @@ Read all about it at http://pitest.org
 
 ### 1.3.3-SNAPSHOT
```
szpak: You may wish to name it 1.4.0 instead of 1.3.3

hcoles: Yes, that was/is the plan

hcoles: now fixed

PR #448, opened 2018-03-01
Adding an additional report-aggregate-module mojo which aggregate using module instead of dependency

sarahBuisson:
@hcoles 
@rchargel

I've slightly modified the new "aggregate-report" mojo class in order to do the aggregation of the report for all the submodule of a project, instead of adding an aggregate project with dependencies.

The configuration of the aggregation will be so much easier and cleaner this way.
 (you can configure everything in the parent project, and you don't have to create a project dedicated to the aggregatation.)

Unfortunately, This method is not compatible with the old report-aggegate mojo, so I had to create a new mojo.

To use-it:
in the pom parent 
```
  <pluginManagement>
                    <plugins>
                        <plugin>
                            <groupId>org.pitest</groupId>
                            <artifactId>pitest-maven</artifactId>
                            <version>${pitest.version}</version>
                            <configuration>
                                <outputFormats>
                                    <outputFormat>XML</outputFormat>
                                </outputFormats>
                                <exportLineCoverage>true</exportLineCoverage>
                  
```
then the command-libe
`
mvn pitest:report-aggregate-module
`


rchargel:
Love it. Thank you. I still owe some documentation

hcoles:
Thanks @sarahBuisson  I will take a look when I get chance, but it will need some integration tests before it can be merged into master.

sarahBuisson:
@hcoles
@rchargel
I've add IT tests,
simple curiosity, do you know when the version 1.4 will be ready?

sarahBuisson:
@hcoles Hello, do you want help with the 1.4? Do you want I do this PR for a previous version of pitest?

hcoles:
@sarahBuisson Thanks for this. The intention is to push 1.4 out when I've had chance to work through the PR backlog - which is proving a little slow as I'm starved of time at the moment.

PR #454, opened 2018-03-29
Fix for the scmRootDir property

giggluigg:
The current production version always sets the scm root dir to the parent basedir, which might not be the correct one in case of a project with multiple levels.
The proposed solution is to have a dedicated property for it, defaulting to the project.parent.basedir, for backward compatibility.
Unfortunately I couldn't figure out a way to create either a Unit or Integration test for it.
Do the same considerations for #409 apply?


giggluigg:
Yes, the commit was pulled in. Although, as I commented on #457, I think we should remove the default value from the annotation, to make it work as expected, if I'm not mistaken. I'll close this one right away, thank you for your time. Sorry for my late reply: had a busy week + national holiday

PR #456, opened 2018-04-02
pitest:scmTestMutationCoverage: scmMutationCoverage but for the modified tests

sarahBuisson:
It's in progress, some unit tests fails, but the principe work.

I've create a new mojo: pitest:scmTestMutationCoverage

The new mojo will do the same things than scmMutationCoverage , but only for the modified test

the consequence is, when I do
` mvn pitest:scmMutationCoverage  pitest:scmTestMutationCoverage -DwithHistory=true `

I will have the mutation for the class who have been modified and all the test, plus the mutation for all the source but only with the modified test.
Aka, everythings we need for launching pitest on a pull-request..

I have a few question:
the issue is, 
` mvn pitest:scmMutationCoverage  pitest:scmTestMutationCoverage -DwithHistory=true  `
is not a command line very easy to understand.

Can I rename the scmMutationCoverage goal into scmSourceMutationCoverage, and re-use the name "scmMutationCoverage" to do a new goal who will do all the work of "mvn pitest:scmMutationCoverage  pitest:scmTestMutationCoverage -DwithHistory=true " ?

### DIFF
--- a/pitest-maven/src/main/java/org/pitest/maven/ScmTestMojo.java
+++ b/pitest-maven/src/main/java/org/pitest/maven/ScmTestMojo.java
@@ -1,0 +1,316 @@
+package org.pitest.maven;
+
+
+import java.io.File;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.function.Predicate;
+
+import org.apache.maven.artifact.Artifact;
+import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.plugins.annotations.Component;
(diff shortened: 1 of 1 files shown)


hcoles:
@sarahBuisson Final had chance to look this over, sorry for the very long delay.

I can see how a goal that gives the mutation coverage of the codebase using only altered tests could be occasionally useful, but I don't follow the logic behind running it in combination with the scm mutation goal

```mvn pitest:scmMutationCoverage pitest:scmTestMutationCoverage```

This will not give a combined report, it will give the report of the last goal to run. I don't think the reports even could be combined as they are showing different things.
 
a) how well tested modified code is by the full test suite
b) how well a the subset of that test suite that has been changed tests the while codebase

I don't see how these two could be logically combined.

Am I missing something?

sarahBuisson:
@hcoles Sorry, it's an old PR I will check that. 
I think I remember you need to add  this  " -DwithHistory=true" in order to have the two report merged.

that the point of my last paragraph : I want to merge all this goal into the same goal, but I need a comprehensible goal name for that.

PR #457, opened 2018-04-02
fix path issue in project with module of module for goal scmMutationCoverage

sarahBuisson:
(again) an issue with the scmMutationCoverage goal for the project with a complexe module hierarchie

The issue was if the projects had more than one level of module ( ie if the module project also had module), the parent-basedir wouldn't match the scm-basedir, so nothing will happens. 
I've create a method to find the good basedir.
Also, depending the directory when you launch maven, the report differ.

I've test the correction on my projects (it work !), but I have trouble with the unit-test of the ScmMojoTest.  If anyone can take a look !


giggluigg:
@sarahBuisson I like your approach, although I see that you removed the property altogether.
In #454 I was biased the other way, by letting the user specify the correct dir and defaulting to the old value. So basically with the other approach you would necessarily have to override the property to make it work with multiple levels. Not the nicest I must admit, just an effective workaround. 
I chose that one because personally, as a user of the feature, I would still like to be able to override the property, to have the highest freedom: what if we join the approaches so that one can specify the property otherwise it defaults to the value computed by your code? I'm not sure it's easy though. @hcoles What's your idea about it?

hcoles:
@sarahBuisson @giggluigg Sorry folks, as you've probably noticed I'm not too responsive at the moment as I have a few other things going on.

I've not been able to look at the details of this PR, but @giggluigg's suggestion of a property that defaults to being automatically computed but can be over-ridden sounds reasonable. If there are reasons why this would be complex or otherwise not viable then if you are able to work together to agree what the best behaviour would be form a user perspective + put together a working PR that would be great.

Thanks

Henry

sarahBuisson:
@hcoles I've rebase the @giggluigg  PR into mine, so the two PR can work together.

 I still have issue with the unit test, thou.  It's hard to build mocks who works with my modif'.

giggluigg:
@hcoles @sarahBuisson Thank you both for taking the time for a feedback.
@sarahBuisson I can make some time later this week to help with the testing part

pitest-maven/src/main/java/org/pitest/maven/ScmMojo.java
```diff
@@ -94,7 +94,7 @@
    * Base of scm root. For a multi module project this is probably the parent
    * project.
    */
-  @Parameter(property = "project.parent.basedir")
+  @Parameter(property = "scmRootDir", defaultValue = "${project.parent.basedir}")
```
giggluigg: Shouldn't we remove the `defaultValue`, to leave it to the code you introduced?

PR #462, opened 2018-04-21
Escape test names when generating coverage xml file

maxgabut:
Test names are usually just method names which can be written in xml without escaping. However, when using Junit's parameterized tests, the test name can be any string the developer of the test choose to use.

Before this commit, if the test name include a quote, the generated xml is not valid.

By using `StringUtil.escapeBasicHtmlChars` on the test name, this commit ensures that any test name can be represented as a valid xml value.

For instance, before this commit, a test named `let's test this!` would have appeared in coverage xml file as `<test name='let's go test this!'/>` which is not valid xml. With this commit it will appear as `<test name='let&#39;s go test this!'/>`.


hcoles:
Thanks!

PR #466, opened 2018-05-05
Ignore absent code directories when aggregating report

maxgabut:
## Current state

Currently, the ReportAggregator builder requires all code directories it works with to be non null, to exist, and to actually be directories. This is standard argument checking and it works just fine for most cases.

However, this behavior is problematic with some maven project structures.

In a maven project using submodules, one may choose to add a non standard source directory. This may be configured in the root pom and applied to all submodules. If this directory has a special purpose, for instance code generated by a given tool, it may not be used by all submodules.

In such a project, when listing source directories, plugins will see the additional source folder everytime, even for submodules where it is not used.

Because of the checks done in ReportAggregator, pitest report aggregation fails when this directory is not present.

## Proposed solution

Some maven plugins works fine in the same situation, for instance, maven-resources-plugin just logs that a given folder was not found and continues its execution.

I propose we do the same here.

### DIFF
--- a/pitest-aggregator/src/main/java/org/pitest/aggregate/ReportAggregator.java
+++ b/pitest-aggregator/src/main/java/org/pitest/aggregate/ReportAggregator.java
@@ -10,6 +10,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.function.Function;
+import java.util.logging.Logger;
 
 import org.pitest.classpath.CodeSource;
 import org.pitest.coverage.BlockCoverage;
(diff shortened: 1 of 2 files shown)


maxgabut:
@hcoles what are you thought on this one ? Do you need me to link this PR to an issue ? (If so I think I can use #467, it's about the problem I'm trying to solve, but in a simpler/clearer form)

hcoles:
@maxgabut Sorry for the slow response - now merged. Thanks for taken the time to create at PR.

Issue #468, opened 2018-05-20
\0 Byte in JUnit tests causes PIT to create invalid XML


uhafner:
In one of my [unit tests](https://github.com/jenkinsci/analysis-model/commit/620477fb2f05ec14e9e9c9a2432857e6b5d9f187) I'm verifying that the 0 byte is correctly handled. While the test passes, PIT (pitest-maven:1.4.0:mutationCoverage) creates an invalid XML file.

``` 
<mutation detected='true' status='KILLED' numberOfTestsRun='4'>
<sourceFile>FingerprintGenerator.java</sourceFile>
<mutatedClass>edu.hm.hafner.analysis.FingerprintGenerator</mutatedClass>
<mutatedMethod>run</mutatedMethod>
<methodDescription>(Ledu/hm/hafner/analysis/FullTextFingerprint;Ledu/hm/hafner/analysis/Report;Ljava/nio/charset/Charset;)V</methodDescription>
<lineNumber>35</lineNumber><mutator>org.pitest.mutationtest.engine.gregor.mutators.VoidMethodCallMutator</mutator>
<index>48</index>
<block>4</block>
<killingTest>edu.hm.hafner.analysis.FingerprintGeneratorTest.[3] Illegal filename = &#0; Null-Byte</killingTest>
<description>removed call to edu/hm/hafner/analysis/Issue::setFingerprint</description>
</mutation>
```

maxgabut:
@uhafner #477 is a way of solving this problem. WDYT ?

uhafner:
Looks good to me, thanks for fixing!

maxgabut:
Great ! I guess you can close the issue now.

hcoles:
@maxgabut Yes, thanks for the PR. I hope to push a release out soon, just need to see if I can find time to look at the other outstanding PR first.

hcoles:
Fixed in 1.4.1

PR #477, opened 2018-06-05
produce valid xml for null-byte in test name

maxgabut:
This solve #468.

A parameterized test name can be anything, and it can even include the null-byte.

It seems that the null-byte is not part of the valid character range for xml (see [Character Range in xml](https://www.w3.org/TR/2006/REC-xml11-20060816/#NT-Char)), so we can not _encode_ it. But nothing prevents us from _representing_ it somehow, and this is what is proposed here.

With this PR, when we need to represent a null-byte in xml, we output `\0` instead. This is not perfect, reversible, a parser would not return a null-byte when reading `\0`, but I believe it to be good enough for reports generation.


hcoles:
Thanks!

PR #478, opened 2018-06-08
Refactor aggregation mojos

maxgabut:
# Current state
There is two mojos for aggregated reports production :
 - `PitAggregationMojo` that select modules to aggregate based on the dependencies of the module it runs in
 - `PitReportAggregationModuleMojo` that is meant to run with command line and select modules to aggregate based on the submodules of the root project

Both mojos share a lot of behavior, some of it by calling the same objects (`ReportAggregator` for instance), some of it by code duplication. This make it complicated to make some changes that should apply to both mojos. (For instance, both mojos can only work if timestamped reports are disabled. Make them accept timestamped reports would require to make changes on both sides)

# Proposed changes
This pull request introduce an abstract class inherited by both mojo classes.

 - `findDependencies` is made abstract, because this is where the difference between both mojos is.
 - `getDescription` is left in both classes. The implementation is currently the same, but I think each implementation could need to define its own name.

# Notable changes
Changes are not easy to spot in moved code, but fortunately few changes where needed.

The only one to notice is in `getCompiledDirs` :

```
-            new PluginServices(PitAggregationMojo.class.getClassLoader())))) {
```
and
```
-            new PluginServices(PitAggregationMojo.class.getClassLoader())))) {
```
where changed to
```
+            new PluginServices(this.getClass().getClassLoader())))) {
```

In context the new version is :
```
+    for (final Object artifactObj : FCollection
+        .filter(project.getPluginArtifactMap().values(), new DependencyFilter(
+            new PluginServices(this.getClass().getClassLoader())))) {
+
+      final Artifact artifact = (Artifact) artifactObj;
+      sourceRoots.add(artifact.getFile().getAbsolutePath());
+    }
```

# Pings (: 
@sbuisson @rchargel What are you thoughts on that ?

### DIFF
--- a/pitest-maven/src/main/java/org/pitest/maven/report/AbstractPitAggregationReportMojo.java
+++ b/pitest-maven/src/main/java/org/pitest/maven/report/AbstractPitAggregationReportMojo.java
@@ -1,21 +1,128 @@
 package org.pitest.maven.report;
 
+import org.apache.maven.artifact.Artifact;
 import org.apache.maven.plugins.annotations.Parameter;
 import org.apache.maven.project.MavenProject;
+import org.apache.maven.reporting.MavenReportException;
+import org.codehaus.plexus.util.FileUtils;
+import org.pitest.aggregate.ReportAggregator;
+import org.pitest.functional.FCollection;
+import org.pitest.maven.DependencyFilter;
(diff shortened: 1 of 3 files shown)


sarahBuisson:
I'm okay with it.

hcoles:
Thanks.

PR #486, opened 2018-07-06
Add synthax highlighting to code in reports

maxgabut:
# Current state
Code in reports is just plain old black on white. It works, but it could be funkier :D

# Proposed change
Add syntax highlighting to the report's code.

## Library choice
I used [code-prettifier](https://github.com/google/code-prettify). I have no strong opinion about this specific lib, I could change it for an other one if it's better somehow.

The lib's useful files are included in the sources and copied to the generated report : this way, it's possible to read the generated report with style event offline.

# Screenshots

## Before

before

## After

after

# Preliminary work
The pull request include some changes that prepared to add syntax highlighting but are not directly related :

## Removing `MutationHtmlReportListener.onRunEnd()`

There seem to have been some confusion between `onRunEnd()` and `runEnd()` : `TestListener` have an `onRunEnd()` method, but `MutationHtmlReportListener` is a `MutationResultListener` which have a `runEnd()` method, it doesn't need to have both.

Since I was about to change `runEnd()` (the "good" one), and `onRunEnd()` had an identical content, I deleted `onRunEnd()` and change the tests that used it to use `runEnd()` assuming the tests really intended to test `runEnd()`.

## Changing file copy mecanism

The only copied file was loaded as a `String` before it was written in a file. Loading it was used at some point to insert the css inside the generated report file. This is not necessary anymore, since the css is now put in a different file.

While doing so, I made a more generic `copyFile` method so it could be used to copy the lib files.


hcoles:
Thanks a lot for this - it will be really good to have syntax hi-lighting.

Just tried this out locally - it looks very nice, except the code-pretifier doesn't seem to understand multiline comments - so it is trying to process them with distracting results.

I'm not familiar with the state of javascript syntax hi-lighters, but might one of he others give better results?

maxgabut:
I think I know why multiline comments are not processed well. In the html report, every line is a `pre` block, and code-pretiffier sees it as an isolated code sample. So when processing a line in the middle of a multi-line comment, it lack context to properly highlight it.

If we can give it the whole code extract at once, it will probably work fine.

maxgabut:
Thinking about it, using an editor like [CodeMirror](https://codemirror.net/) in read only mode is probably the way to go.

It's possible to put [almost anything in the gutter](https://codemirror.net/demo/lint.html) (hover over markers to see how cool they are). It's also possible to add info under lines [like here](https://codemirror.net/demo/widget.html). I think its possible to make something really rich.

There is a problem though, to do this kind of development I think it should be possible to iterate really fast, and running the whole analysis to generate an html page is too long (or short but with very few data). So I think that we should first have a way to dump all the data we need in a file (including source files) and then have a js app read this file to display its content.

I'm closing this PR, because the new proposition will be really different (and won't be ready for while), but if anyone has ideas on that I'd be glad to here about it.

hcoles:
Thanks - I look forward to the PR, it would be really good to spruce up the report.

Issue #490, opened 2018-07-25
Argument list too long for minions on long class paths


davidburstromspotify:
In case the test class path is very long, there is an option to supply the class path through a file `classPathFile`. However, we still run into problems starting the minions, because internally, Pitest doesn't seem to use a file to supply the class path to them:

```
Exception in thread "main" org.pitest.util.PitError: Cannot run program "/usr/lib/jvm/jdk-8-oracle-x64/jre/bin/java": error=7, Argument list too long
at org.pitest.util.Unchecked.translateCheckedException(Unchecked.java:20)
	at org.pitest.coverage.execute.DefaultCoverageGenerator.calculateCoverage(DefaultCoverageGenerator.java:103)
	at org.pitest.coverage.execute.DefaultCoverageGenerator.calculateCoverage(DefaultCoverageGenerator.java:49)
	at org.pitest.mutationtest.tooling.MutationCoverage.runReport(MutationCoverage.java:115)
	at org.pitest.mutationtest.tooling.EntryPoint.execute(EntryPoint.java:115)
	at org.pitest.mutationtest.tooling.EntryPoint.execute(EntryPoint.java:49)
	at org.pitest.mutationtest.commandline.MutationCoverageReport.runReport(MutationCoverageReport.java:87)
	at org.pitest.mutationtest.commandline.MutationCoverageReport.main(MutationCoverageReport.java:45)
Caused by: java.io.IOException: Cannot run program "/usr/lib/jvm/jdk-8-oracle-x64/jre/bin/java": error=7, Argument list too long
	at java.lang.ProcessBuilder.start(ProcessBuilder.java:1048)
	at org.pitest.process.WrappingProcess.start(WrappingProcess.java:45)
	at org.pitest.coverage.execute.CoverageProcess.start(CoverageProcess.java:30)
	at org.pitest.coverage.execute.DefaultCoverageGenerator.gatherCoverageData(DefaultCoverageGenerator.java:131)
	at org.pitest.coverage.execute.DefaultCoverageGenerator.calculateCoverage(DefaultCoverageGenerator.java:87)
	... 6 more
Caused by: java.io.IOException: error=7, Argument list too long
	at java.lang.UNIXProcess.forkAndExec(Native Method)
	at java.lang.UNIXProcess.<init>(UNIXProcess.java:247)
	at java.lang.ProcessImpl.start(ProcessImpl.java:134)
	at java.lang.ProcessBuilder.start(ProcessBuilder.java:1029)
	... 10 more
```

PR #496, opened 2018-08-06
Add support for passing long class paths to minions

jqhan:
This is a fix for this [issue](https://github.com/hcoles/pitest/issues/490)

> In case the test class path is very long, there is an option to supply the class path through a file classPathFile. However, we still run into problems starting the minions, because internally, Pitest doesn't seem to use a file to supply the class path to them

This solution will pack the class path into a jar file, that will be sent to the minions, whenever a class path **file** has been supplied to the MutationCoverage process. The class path jar file gets created in the same way as it does in Intellij. Here's some more background information about the problem: https://blog.jetbrains.com/idea/2017/10/intellij-idea-2017-3-eap-configurable-command-line-shortener-and-more/

### DIFF
--- a/pitest-command-line/src/main/java/org/pitest/mutationtest/commandline/OptionsParser.java
+++ b/pitest-command-line/src/main/java/org/pitest/mutationtest/commandline/OptionsParser.java
@@ -434,6 +434,7 @@ public class OptionsParser {
           ClassPath.getClassPathElementsAsPaths(), this.dependencyFilter));
     }
     if (userArgs.has(this.classPathFile)) {
+      data.setShouldUseClassPathJar(true);
       try (BufferedReader classPathFileBR = new BufferedReader(new FileReader(userArgs.valueOf(this.classPathFile).getAbsoluteFile()))) {
         String element;
         while ((element = classPathFileBR.readLine()) != null) {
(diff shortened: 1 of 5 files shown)


hcoles:
Thanks for this. I've had a play in a branch over here

hcoles:
Thanks for this, I've had a play in a branch over here

https://github.com/hcoles/pitest/tree/jqhan-classpath-file-arg

Which always uses a classpath jar. This flushed out a few other changes that were required so that pitest would work correctly when the classpath is passed in this way.

This now works, but your link suggests some libraries and frameworks may have issues with a classpath jar.

So this may need to be made properly configurable and options added to the various build integrations.

However a quick google has suggested a simple alternative. On modern versions of windows this link suggests simply passing the classpath as an environment variable bypasses length limits.

https://issues.apache.org/jira/browse/SUREFIRE-727

Unfortunately I do not have access to a windows machine to test this out.

jqhan:
Yes, this now works with a classpath jar. It would be great if someone could implement an alternative which provides the classpath as a text file, as it was mentioned in the Intellij link.

I'm not sure about how the ProcessBuilder works under the hood when it's setting the CLASSPATH environment variable, but from what I've experienced the processBuilder seems to eventually put the whole classpath on the command line which causes the "Argument list too long" error. Prior to this PR, passing the classpath by setting the processBuilder's CLASSPATH environment variable was the only alternative, and the error still occurred.

By the way, I think you introduced a bug in the LaunchOptions constructor. The usingClassPathJar instance variable shouldn't always be set to false no matter what, right?

jqhan:
Just a question, when do you think that this fix will be released? We're currently blocked from using pitest on a large part of our code base because of this problem :(

hcoles:
@jqhan since it was blocking you I cut a release. 1.4.2 should now have synced to maven central.

PR #501, opened 2018-08-10
Update LaunchOptions.java

jqhan:
Instance variable shouldn't always be false


hcoles:
Thanks - clearly this is in need of an integration test.

PR #505, opened 2018-08-30
Add LGTM code quality badges

xcorail:
Hi there!

I thought you might be interested in adding these [code quality badges](https://lgtm.com/projects/g/hcoles/pitest/ci/#badges) to your project. They will indicate a very high code quality to potential users and contributors.
You can also check the alerts discovered by [LGTM](https://lgtm.com/projects/g/hcoles/pitest).

N.B.: I am on the team behind LGTM.com, I'd appreciate your feedback on this initiative, whether you're interested or not, if you find time to drop me a line. Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 [![Maven Central](https://maven-badges.herokuapp.com/maven-central/org.pitest/pitest/badge.svg?style=flat)](https://maven-badges.herokuapp.com/maven-central/org.pitest/pitest)
 [![Build Status](https://travis-ci.org/hcoles/pitest.png?branch=master)](https://travis-ci.org/hcoles/pitest)
+[![Code Quality: Java](https://img.shields.io/lgtm/grade/java/g/hcoles/pitest.svg?logo=lgtm&logoWidth=18)](https://lgtm.com/projects/g/hcoles/pitest/context:java)
+[![Total Alerts](https://img.shields.io/lgtm/alerts/g/hcoles/pitest.svg?logo=lgtm&logoWidth=18)](https://lgtm.com/projects/g/hcoles/pitest/alerts)
 
 Pitest (aka PIT) is a state of the art mutation testing system for Java and the JVM.
 


maxgabut:
I had a quick look at the reports. I think it's a good idea to do code analysis as a service, and that it is quite nice to do it for free for open source projects.

However, for the whole code base, LGTM only found 3 alerts. It looks _really_ low to me. I ran a sonarcloud analysis and it found 51 bugs, 6 vulnerabilities and 484 code smells.

Maybe something is going wrong during the LGTM analysis ?

Anyway, with the type of results we currently get, I don't think a LGTM badge is really meaningful.

xcorail:
Hello @maxgabut 

> I had a quick look at the reports. I think it's a good idea to do code analysis as a service, and that it is quite nice to do it for free for open source projects.

Thanks for the feedback! 

> However, for the whole code base, LGTM only found 3 alerts. It looks really low to me. I ran a sonarcloud analysis and it found 51 bugs, 6 vulnerabilities and 484 code smells.

It can be that something went wrong, can you please give me a few examples (just 2 or 3) of the 51 bugs or 6 vulnerabilities, that you actually consider as real problems in your code, and that we didn't catch, and I will have a look with my Java team. 
 
But it can be also because LGTM does not display all alerts by default. Indeed, 
* We pay a lot of attention to avoid displaying false positives, so we display only the results of our high precision queries (e.g. a very low rate of false positives)
* We also don't want to flood a developer with 500 issues that he won't fix. So we analyze the behaviour of developers by data mining about 20 millions commits on GitHub, and we know what alerts developers fix, and what alerts they don't. And the alerts that developers don't seem to care about (perhaps your Sonarcloud issues are in this case?), we don't display them by default. 

You can override this default and display all results as described in the [help](https://lgtm.com/help/lgtm/showing-hiding-query-results), you can add a `lgtm.yml` file like below and wait for the next analysis (LGTM polls your repo once a day)
```
queries:
    - include: "*"
```

> I don't think a LGTM badge is really meaningful.

The service is free, even without the badge. It's more important for me to make it meaningful as you said, so if you have time to give me these examples I can dig in and find out.  

Cheers, thanks again for your time and your feedback.

hcoles:
Hi @xcorail

Thank you for the pull request, but I do not think this is the way we wish to improve the pitest codebase.

PR #506, opened 2018-08-31
Improve the logging in case of a non-green test suite

nrainer:
... and show the number of failing tests


nrainer:
Replaced by #507.

PR #507, opened 2018-08-31
Improve the logging in case of a non-green test suite 

nrainer:
... and show the number of failing tests

### DIFF
--- a/pitest-entry/src/main/java/org/pitest/coverage/CoverageData.java
+++ b/pitest-entry/src/main/java/org/pitest/coverage/CoverageData.java
@@ -57,7 +57,7 @@ public class CoverageData implements CoverageDatabase {
 
   private final LineMap                                       lm;
 
-  private boolean                                             hasFailedTest = false;
+  private int                                                 countFailedTests = 0;
 
   public CoverageData(final CodeSource code, final LineMap lm) {
     this(code, lm, new LinkedHashMap<BlockLocation, Set<TestInfo>>());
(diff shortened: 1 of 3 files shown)


hcoles:
There look to be some issues with travis at the moment, don't look to be related to the change.

I forget how this works, but I think pitest might abort as soon as it hits a failing test. Do you ever see counts > 1 where there are failures in different test classes? (it might not be able to abort if it is running tests in the same atomic unit).

nrainer:
Hi Henry,
thanks for your comment. :)

If tried this out and it looks to me as if in the initial coverage analysis all tests get executed. However, I am not fully sure about it. I tried it out with PIT 1.3.2, but I have at the same time a patched version of PIT 1.4.0 installed with modifications to the MutationTestWorker to create a full mutation matrix (see commit https://github.com/nrainer/pitest/commit/610fb9d35596017455c50e5d1819ab98cfedb71b). Though, I would assume that this does not influence the result.

(By the way, would it be imaginable to integrate the full mutation matrix feature as configuration option? If so, I would be happy to provide a pull request.)

hcoles:
@nrainer Could you confirm that the change works as expected in this branch of the code?

I'd be happy to accept a PR that implemented a matrix cleanly, unfortunately this is not straight forward as it has implications for the report which currently has no way of representing the results of a matrix.

nrainer:
@hcoles I retested it with 1.4.2 and it looks good to me. No early test exit on failures occurs during initial coverage collection and the changes work as expected. Therefore, in my opinion, the pull request can be merged.

Regarding the mutation matrix: Ok, I will prepare a pull request as discussion basis. :)

PR #508, opened 2018-08-31
Html generation minor improvements

maxgabut:
This pull request fixes two small issues I noticed regarding html report.

1) Relocating templates

Templates and css files used for html report generation were located in the `pitest` maven module.
Since those files are only used by the `pitest-maven-report`, I moved them to this module where I think they belong.

2) Closing span

The generated html files contained an unclosed span. I closed it.


szpak:
@maxgabut Is the first point related to multimodule report generation? If yes, it will be probably used by the Gradle module (once [implemented](https://github.com/szpak/gradle-pitest-plugin/issues/80)).

maxgabut:
Well, it is related, but depending on the way the gradle plugin works, it can be ok. The dependencies between the maven module of interest here is this way :

```
[pitest] <- [pitest-html-report] <- [pitest-aggregator] <- [pitest-maven]
```
`pitest-html-report` uses templates to generate reports
`pitest-aggregator` uses individual reports and `pitest-html-report` to generate an aggregated report
`pitest-maven` uses `pitest-aggregator` to drive `pitest-aggregator` as a maven plugin

When implementing the multimodule report generation in the Gradle plugin, I think dependencies will look that way :

```
[pitest] <- [pitest-html-report] <- [pitest-aggregator] <-+---- [pitest-maven]
                                                          +---- [pitest-gradle]
```

So I think moving those files from `pitest` to `pitest-html-report` will not be a problem for that.

szpak:
I think you are right. The diff view in GitHub shrinks the paths, so I wasn't sure where you want to move it. I asked the author of that feature during development to keep it possibly separate from the pitest-maven module, so I expect it should be achievable to reuse the logic.

PR #516, opened 2018-09-14
Log number of processed mutation units

nrainer:


maxgabut:
@hcoles what do you think of this one ? It's simple and low risk.

It's not useful in all use case and all projects structure, but... I think it's always nice for long running tools to log something once in a while. At least you know that it's still running and making progress.

PR #518, opened 2018-09-18
BigInteger mutator

ripdajacker:
My colleague and I added a new experimental mutator that mutates BigInteger methods in the same manner as the MathMutator does for ints and longs. 

The mutator also swaps methods that are not present on the primitives:

* andNot <> and
* setBit <> clearBit
* flipBit > setBit
* not <> negate
* max <> min
* abs > negate

We think this is a useful addition and would love to hear some feedback. 

Note: The branch has BigDecimal in the name, but BigDecimal is much more complicated so we opted to do a test with BigInteger first.


hcoles:
Thanks for this - finally got chance to have a look over it yesterday.

I agree this would be good to add in. In the longer term I think it might be possible to replace this with a more generalised method swap mutator, but that will require some changes to pitest to implement so it unlikely to happen soon.

If you can fill in the holes in the test suite (at the moment the tests only demonstrate swapping a subset of the subsituted methods) I'll merge it in.

ripdajacker:
We added tests for the rest of the methods.

hcoles:
Thanks! It is not accessible as EXPERIMENTAL_BIG_INTEGER in the master branch

PR #523, opened 2018-10-03
sorted mutators in html report

ThLeu:
Hi!

This is a very minor enhancement of the HTML report.
When there are many mutators active, especially the experimental remove switch mutator than the active mutators list gets kind of messy.
So I changed the HashSet to a TreeSet to get a better overview of them.

Btw. if you agree I would also work on a little enhancement to aggregate mutators like the EXPERIMENTAL_REMOVE_SWITCH_MUTATOR_X to something like 
`EXPERIMENTAL_REMOVE_SWITCH_MUTATOR_[1-99]` in the active mutator list.

grafik


hcoles:
thanks!

PR #524, opened 2018-10-04
provide aggregated name for mutator if generated as a collection

ThLeu:
Hi, as I mentioned in my other PR here is also some minor enhancement to aggregate the EXPERIMENTAL_REMOVE_SWITCH_MUTATOR in the report.

Best regards,
ThLeu

### DIFF
--- a/pitest/src/main/java/org/pitest/mutationtest/engine/gregor/mutators/experimental/RemoveSwitchMutator.java
+++ b/pitest/src/main/java/org/pitest/mutationtest/engine/gregor/mutators/experimental/RemoveSwitchMutator.java
@@ -22,21 +22,15 @@ public class RemoveSwitchMutator implements MethodMutatorFactory {
   private static final int GENERATE_UPTO_EXCLUDING = 100;
 
   private final int key;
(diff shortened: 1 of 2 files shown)


ThLeu:
Hello @hcoles!
What do you think of this change? Would appreciate your feedback.
Best regards

hcoles:
Thanks for the PR.

Just had a quick scan over - what's the purpose of the `isGeneratedAsCollection` flag? It appears that this will always be set to true, apart from in tests.

hcoles:
But yes, since the mutation description includes info about which switch variation was performed there is no need for it to be in the mutator name and this is definitely a big improvement - thanks.

ThLeu:
Thanks for your feedback!
The `isGeneratedAsCollection` is only there because the constructor is public, so I thought that it could and maybe will be used by clients somehow. As alternative the constructor could be set to default visibility (package private) if it is only used in the test, then the flag can be deleted again.

ThLeu:
I changed it now to the mentioned version, if it does not matter that the constructor is package private than this version is probably better.

PR #528, opened 2018-10-16
Added skipFailingTests option from maven plugin

nicerloop:
Added an option to ignore failing tests from coverage, activated from maven plugin

### DIFF
--- a/pitest-entry/src/main/java/org/pitest/coverage/execute/DefaultCoverageGenerator.java
+++ b/pitest-entry/src/main/java/org/pitest/coverage/execute/DefaultCoverageGenerator.java
@@ -179,7 +179,9 @@ public class DefaultCoverageGenerator implements CoverageGenerator {
 
       @Override
       public void apply(final CoverageResult cr) {
-        coverage.calculateClassCoverage(cr);
+        if (cr.isGreenTest() || !coverageOptions.getPitConfig().skipFailingTests()) {
+          coverage.calculateClassCoverage(cr);
+        }
         if (DefaultCoverageGenerator.this.showProgress) {
           System.out.printf("%s", this.spinner[this.i % this.spinner.length]);
         }
--- a/pitest-maven/src/main/java/org/pitest/maven/SurefireConfigConverter.java
+++ b/pitest-maven/src/main/java/org/pitest/maven/SurefireConfigConverter.java
@@ -81,10 +81,9 @@ public class SurefireConfigConverter {
   }
 
   private void convertTestFailureIgnore(ReportOptions option, Xpp3Dom configuration) {
-    List<String> testFailureIgnoreList = extractStrings("testFailureIgnore", configuration);
-    if (!testFailureIgnoreList.isEmpty()) {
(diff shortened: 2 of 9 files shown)


pitest-entry/src/main/java/org/pitest/coverage/execute/DefaultCoverageGenerator.java
```diff
@@ -179,7 +179,9 @@ private void gatherCoverageData(final Collection<ClassInfo> tests,
 
       @Override
       public void apply(final CoverageResult cr) {
-        coverage.calculateClassCoverage(cr);
+        if (cr.isGreenTest() || !coverageOptions.getPitConfig().skipFailingTests()) {
```
pedrorijo91: should we at least log that we are skipping coverage computation due to flag+failed tests?

pitest-maven/src/main/java/org/pitest/maven/SurefireConfigConverter.java
```diff
@@ -79,4 +80,11 @@ private void convertExcludes(ReportOptions option, Xpp3Dom configuration) {
     return Collections.emptyList();
   }
 
+  private void convertTestFailureIgnore(ReportOptions option, Xpp3Dom configuration) {
+    List<String> testFailureIgnoreList = extractStrings("testFailureIgnore", configuration);
+    if (!testFailureIgnoreList.isEmpty()) {
```
pedrorijo91: what happens if the list contains more than 1 element? is it ignored? should we log?

nicerloop: I could just skip the list conversion and directly extract the value.

pitest-entry/src/main/java/org/pitest/coverage/execute/DefaultCoverageGenerator.java
nicerloop: I actually don't know the strategy around logs in this project, just hacked for my immediate need to skip failing tests

nicerloop: Where in the code can I see the recommended pattern to add logging?

nicerloop:
Is this evolution in the line of the project? May it be integrated? What would be required for that?

hcoles:
@nicerloop sorry for the slow response, unfortunately I'm a little short of time these days.

Skipping failing tests has been requested as a feature a few times now, I've always been reluctant to implement it as it adds one more variable to consider when people raise an issue because they believe there is a falsely killed mutant. If a failing test is not properly excluded from running it will appear to kill all mutants. 

Not sure if this is still a rational concern or not: I'll take a look at the PR when I get chance.

dormaayan:
Thanks for implementing this! It seems useful!
Is there a plan to add this feature to the upcoming release of Pitest?

maxgabut:
From the merge date, it's included in version 1.4.4.

It's lacking documentation at that point, but adding `<skipFailingTests>true</skipFailingTests>` in the plugin configuration in you pom should do the trick (not tested, but from the code it should work)

PR #532, opened 2018-11-04
Add some rules regarding issues management

maxgabut:
Hi @hcoles, as discussed together, this is first draft for a set of rules we could apply to be able to work more efficiently with pitest issues.

Anyone should feel free to comment, and propose changes. Once we agree, we can start applying those rules.

The main point of this PR is the `ISSUES.md` file. I also made some changes in `README.md` : in order to have a link to the issue policy that's visible enough I folded most of the versions list. This is easy to undo if needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -386,7 +386,7 @@ however disabled by default, and the generated errors are correctly handled by P
 
 ## Issues
 
-Issues management is described [here](ISSUES.md).
+Please consult our [issue management rules](ISSUES.md) before creating or working on issues. 
 
 ## Credits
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -379,6 +382,11 @@ however disabled by default, and the generated errors are correctly handled by P
 ### 0.18
 
 * First public release
+</details>
+
+## Issues
+
+Issues management is described [here](ISSUES.md).
```
pzygielo: Only one click away from [Don't use "click here" as link text](https://www.w3.org/QA/Tips/noClickHere).

maxgabut: :sweat_smile: I rephrased it : what do you think about the new version ?

hcoles:
Thanks @maxgabut, looks good.

PR #533, opened 2018-11-05
Add classic operators from the litterature

LaurentTho3:
Add the operators introduced in https://researchrepository.ucd.ie/bitstream/10197/7748/4/ISSTA_2016_Demo_Camera_ready.pdf to pitest.

### DIFF
--- a/pitest/src/main/java/org/pitest/mutationtest/engine/gregor/config/Mutator.java
+++ b/pitest/src/main/java/org/pitest/mutationtest/engine/gregor/config/Mutator.java
@@ -49,6 +49,7 @@ import org.pitest.mutationtest.engine.gregor.mutators.RemoveConditionalMutator;
 import org.pitest.mutationtest.engine.gregor.mutators.RemoveConditionalMutator.Choice;
 import org.pitest.mutationtest.engine.gregor.mutators.ReturnValsMutator;
 import org.pitest.mutationtest.engine.gregor.mutators.VoidMethodCallMutator;
+import org.pitest.mutationtest.engine.gregor.mutators.experimental.ABSMutator;
 import org.pitest.mutationtest.engine.gregor.mutators.experimental.AOR1Mutator;
 import org.pitest.mutationtest.engine.gregor.mutators.experimental.AOR2Mutator;
 import org.pitest.mutationtest.engine.gregor.mutators.experimental.AOR3Mutator;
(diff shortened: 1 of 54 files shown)


hcoles:
Thanks @LaurentTho3 I'll take a look over this when I get chance, which might take a little while.

maxgabut:
It might also be nice to document those new mutators in the [pitest site](https://github.com/hcoles/pitest-site).

pitest/src/main/java/org/pitest/mutationtest/engine/gregor/config/Mutator.java
```diff
@@ -177,11 +202,72 @@
...
+    add("EXPERIMENTAL_OBBN3", OBBN3Mutator.OBBN_3_MUTATOR);
+
+    /**
+     * Experimental mutators that replace conditional operators.
+     */
+    add("EXPERIMENTAL_ROR1", ROR1Mutator.ROR_1_MUTATOR);
+    add("EXPERIMENTAL_ROR2", ROR2Mutator.ROR_2_MUTATOR);
+    add("EXPERIMENTAL_ROR3", ROR3Mutator.ROR_3_MUTATOR);
+    add("EXPERIMENTAL_ROR4", ROR4Mutator.ROR_4_MUTATOR);
+    add("EXPERIMENTAL_ROR3", ROR5Mutator.ROR_5_MUTATOR);
```
ayaankazerouni: Should this be `"EXPERIMENTAL_ROR5"`?

```suggestion
    add("EXPERIMENTAL_ROR5", ROR5Mutator.ROR_5_MUTATOR);
```

LaurentTho3: It should, indeed. Thanks!

LaurentTho3:
Hello @hcoles ,
UOI1 and UOI2 replace a by a++ and a--, i.e they replace
```
FLOAD var
```
by 
```
FLOAD var
FDUP
FCONST 1
FADD / FSUB
FSTORE var
```
So if the first FLOAD is not there they mutate "return f;" to "return f++/--;", which is an equivalent mutant. So g is only there for the mutant to be detectable. UOI3 and UOI4 that replace a by ++a and --a do not have that g variable in their tests.

I hope that makes sense.

ayaankazerouni:
Just out of curiosity:

Jeff Offutt's original sufficient set also included the `LCR` operator, which "replaces each logical connector (AND and OR) with several kinds of logical connectors" [\[1\]](https://dl.acm.org/citation.cfm?id=227610).

@LaurentTho3 what is the reason that wasn't implemented as part of this PR? Did it get folded into one of the other operators? Or was there some PIT-specific reason there isn't a clear mapping for that operator (e.g. because PIT operates on bytecode)?

Sorry if this is an obvious question, but I don't have an in-depth understanding of Java bytecode. Thanks!

LaurentTho3:
Hi @ayaankazerouni,

AND and OR are not translated directly into a single bytecode instruction each, but are decomposed into a series of branching instructions (ifeq and ifne basically).

Because of this, && and || are mutated by ROR, which will mutate the ifeq and ifne instructions. These mutants might be a bit hard to understand (see https://groups.google.com/d/topic/pitusers/dZrOBoeHgs4/discussion) and might even not map straight to source code mutants (I have not looked into that). Making an operator that directly replaces && with || or vice versa would be complicated (not possible?) with PIT, I think.

PR #544, opened 2018-12-22
GregorMutater line 98 Minor Bug Fix

ali-ghanbari:
GregorMutater line 98 changed from asJavaName to asInternalName so that we can take advantage of caching of ByteArraySource.


maxgabut:
Hi @ali-ghanbari, thanks for the pull request.

Looking at your code, I can't figure out exactly what bug you are fixing. Can you provide a description of the bug and steps to reproduce it ?

Issue #539, opened 2018-12-08
Glob NullPointerException


rnveach:
Accidentally adding a empty `TargetTest` results in an NPE.

Example:
````
              <targetTests>
                <param>com.puppycrawl.tools.checkstyle.xpath.*</param>
                <param>com.puppycrawl.tools.checkstyle.XpathFileGeneratorAuditListenerTest</param>
                <param>com.puppycrawl.tools.checkstyle.XpathFileGeneratorAstFilterTest</param>
                <param></param>
              </targetTests>
````

Trace:
````
[ERROR] Failed to execute goal org.pitest:pitest-maven:1.4.3:mutationCoverage (default-cli) on project checkstyle: Execution default-cli of goal org.pitest:pitest-maven:1.4.3:mutationCoverage failed. NullPointerException -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.pitest:pitest-maven:1.4.3:mutationCoverage (default-cli) on project checkstyle: Execution default-cli of goal org.pitest:pitest-maven:1.4.3:mutationCoverage failed.
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:224)
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:153)
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:145)
	at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:116)
	at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:80)
	at org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build(SingleThreadedBuilder.java:51)
	at org.apache.maven.lifecycle.internal.LifecycleStarter.execute(LifecycleStarter.java:120)
	at org.apache.maven.DefaultMaven.doExecute(DefaultMaven.java:355)
	at org.apache.maven.DefaultMaven.execute(DefaultMaven.java:155)
	at org.apache.maven.cli.MavenCli.execute(MavenCli.java:584)
	at org.apache.maven.cli.MavenCli.doMain(MavenCli.java:216)
	at org.apache.maven.cli.MavenCli.main(MavenCli.java:160)
[...]

PR #555, opened 2019-01-20
Ignore empty element from plugin configuration

maxgabut:
As highlighted by https://github.com/hcoles/pitest/issues/539, throwing a `NullPointerException` when the plugin configuration contains an empty element is not the most user friendly behavior.

This pull request change `AbstractPitMojo` getters to filter out nulls from lists. This way, empty elements in configuration will be ignored.

Before the actual bug fix, fields were made private and getters and setters were created as needed.

*PR history:*
A first version of this PR replaced `ArrayList` fields and return types by `List`. This was removed to keep compatible with maven 2.

### DIFF
--- a/pitest-maven/src/main/java/org/pitest/maven/AbstractPitMojo.java
+++ b/pitest-maven/src/main/java/org/pitest/maven/AbstractPitMojo.java
@@ -6,6 +6,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.logging.Logger;
 
 import org.apache.maven.artifact.Artifact;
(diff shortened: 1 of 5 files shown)


hcoles:
The use of ArrayLists is intentional - earlier versions of maven throw an error if you use an abstract list type.

I'm not sure at what point this was fixed, but the problem is certainly present in 2.0.9.

maxgabut:
I understand that a tool like Pitest needs to accept old versions of all its dependencies, but maven 2 is getting [seriously old](https://maven.apache.org/docs/history.html). Since it reached end of life almost five years ago, I think that we could safely drop support for it.

However, if you think that we still need it, I'll update this PR to leave the `ArrayList` as they are (even if it's [bad for fairies](https://stackoverflow.com/questions/17796756/maven-2-reports-java-util-list-cannot-be-instantiated-error) :wink: )

maxgabut:
@hcoles I dropped to removal of the `ArrayList`.

Issue #542, opened 2018-12-10
Mutator enumation name doesn't match mutation in configs


rnveach:
https://github.com/hcoles/pitest/blob/6101377de641c57a17ff82545682b9b7fbc3c061/pitest/src/main/java/org/pitest/mutationtest/engine/gregor/config/Mutator.java#L90

Enumeration is called `VOID_METHOD_CALL_MUTATOR` and name in configuration is `VOID_METHOD_CALLS`.

The enumeration is missing the `S` on the end of `CALL`.
I noticed this because I was going by the code's default list to copy into the config and this one failed.

PR #558, opened 2019-02-02
Improve naming consistency in Mutator

maxgabut:
# Current state
Issue #542 pointed out some differences between Mutator constants and their names in configuration. While this does not prevent anything from working, it can make the code harder to read ; one expects the 'S' in the mutation name to always appear in the constant name (or never, but not "some time").

# Proposed changes
In this pull request, I tried to increase the consistency of classes, constants and method naming as they appear in the `Mutator` class.

# Reviewing this
A lot of files where changed, but all changes are the consequences of refactoring made from the `Mutator` class. This is the first class to review to get a sense of the changes. Finally, I tried to group similar changes by commit, so if one type of change is not wanted, it should be easy to revert some and to keep the rest.


maxgabut:
@hcoles any thoughts on this one ?

Issue #564, opened 2019-02-12
Windows line endings in repo


rnveach:
I assume this repo should have full linux line endings.
````
$ find . -not -type d -exec file "{}" ";" | grep CRLF

./pitest-maven-verification/src/test/resources/pit-33-setUserDir/pit-33-setUserDir-subModule/pom.xml: ASCII text, with CRLF line terminators
./pitest-maven-verification/src/test/resources/pit-33-setUserDir/pit-33-setUserDir-subModule/src/test/java/org/example/FileOpeningTest.java: ASCII text, with CRLF line terminators
./pitest-maven-verification/src/test/resources/pit-33-setUserDir/pit-33-setUserDir-subModule/src/main/java/org/example/SystemUnderTest.java: ASCII text, with CRLF line terminators
./pitest-ant/src/test/java/org/pitest/ant/PitestTaskTest.java: ASCII text, with CRLF line terminators
./pitest-ant/src/main/java/org/pitest/ant/PitestTask.java: ASCII text, with CRLF line terminators
./pitest/src/test/java/org/pitest/util/MemoryMonitorTest.java: ASCII text, with CRLF line terminators
./pitest/src/test/java/org/pitest/mutationtest/engine/gregor/mutators/experimental/MutantStarter.java: ASCII text, with CRLF line terminators
./pitest/src/test/java/org/pitest/mutationtest/engine/gregor/mutators/experimental/ReturnValuesMutatorTest.java: ASCII text, with CRLF line terminators
./pitest/src/test/java/org/pitest/mutationtest/engine/gregor/mutators/experimental/MemberVariableMutatorTest.java: ASCII text, with CRLF line terminators
./pitest/src/test/java/org/pitest/mutationtest/engine/gregor/mutators/experimental/MutantStarterTest.java: UTF-8 Unicode text, with CRLF line terminators
./pitest/src/test/java/org/pitest/mutationtest/engine/gregor/mutators/experimental/SwitchMutatorTest.java: ASCII text, with CRLF line terminators
./pitest/src/test/java/org/pitest/mutationtest/engine/gregor/mutators/experimental/RemoveSwitchMutatorTest.java: ASCII text, with CRLF line terminators
./pitest/src/test/java/org/pitest/mutationtest/engine/gregor/mutators/RemoveConditionalMutatorTest.java: ASCII text, with CRLF line terminators
[...]

PR #570, opened 2019-02-15
Only accept LF (linux style) end of lines

maxgabut:
There is currently multiple style of line ending in the project.
As suggested in https://github.com/hcoles/pitest/issues/564, I think that there should be only one style in the code base.

In this PR I added a checkstyle rule to only allow linux style end of line and I changed the files that had different style.

**review note:** the diff is quite heavy, but ignoring whitespace changes will remove all changes except the checkstyle rule.

@hcoles WDYT ?


rnveach:
@maxgabut you'll have to rebase this because of changes made in a PR I made that got accepted.

hcoles:
@maxgabut Thanks for this, sorry for the slow response. This will get merged in after the changes to the coverage system in #534 (it's a large change + complex change and I'd prefer not to have to fight through diffs if any issues occur).

PR #576, opened 2019-02-24
Added call to .close() for ReadableByteChannel, WritableByteChannel and PrintWriter

Michael-Nolan:
Added implicit call to .close by using try w resources.


hcoles:
@Michael-Nolan 

Thanks for the PR, however the writers are owned by the factory that produces them. Responsibility for closing them lies there. This is triggered by a call to `close` on the factory in a finally block within the EntryPoint class.

Closing the writers elsewhere will likely to break functionality as data is written to the output file in multiple stages.

Michael-Nolan:
Thank you for the feedback.

PR #590, opened 2019-03-28
add windows to build maatrix

hcoles:


pitest/src/test/java/org/pitest/util/ManifestUtilsTest.java
```diff
@@ -11,7 +11,7 @@
 
   @Test
   public void shouldParseTheManifestsItCreates() throws IOException {
-    File actual = ManifestUtils.createClasspathJarFile("/some/path/foo.jar:/some/path/");
+    File actual = ManifestUtils.createClasspathJarFile("/some/path/foo.jar" + File.pathSeparator +"/some/path/");
     assertThat(ManifestUtils.readClasspathManifest(actual))
     .containsExactly(new File("/some/path/foo.jar"), new File("/some/path/"));
```
rnveach: > Expecting:
  <[D:\some\path\foo.jar, D:\some\path]>
to contain exactly (and in same order):
  <[D:\some\path\foo.jar, D:\some\path]>
but some elements were not found:
  <[D:\some\path\foo.jar, D:\some\path]>
and others were not expected:
  <[D:\some\path\foo.jar, D:\some\path]>

When I run this locally actual is like `C:\some\...` while expected is like `\some\...`. I believe the output is slightly misleading in what the failure is.

The utility is telling if the objects match by using `Objects.areEqual(..., ...)` which is just basically `o1.equals(o2)`

The test passes for me when I change the code to:
````
.containsExactly(new File(new File("/some/path/foo.jar").getAbsolutePath()), new File(new File("/some/path/").getAbsolutePath()));
````
I didn't test how linux would respond.

PR #606, opened 2019-05-08
Update so_you_want_to_build_mutation_testing_system.md

errandir:
Minor fixes and enhancements.

### DIFF
--- a/so_you_want_to_build_mutation_testing_system.md
+++ b/so_you_want_to_build_mutation_testing_system.md
@@ -2,7 +2,7 @@
 
 ## Introduction
 
-There have been a lot mutation testing systems, but very few have them have seen successful use in industry.
+There have been a lot mutation testing systems, but very few of them have seen successful use in industry.
 
 This document is a set of notes that might be helpful for anyone thinking of implementing a mutation testing system for another language.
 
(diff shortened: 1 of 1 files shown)


so_you_want_to_build_mutation_testing_system.md
```diff
@@ -109,7 +109,7 @@ Academic systems tend to keep running so that a matrix can be produced of which
 
 There is a lot of complicated code in pitest that splits tests up into the smallest possible individually executable units. The default in Java is for all the tests defined in single test to be run as one (at least for JUnit and TestNG).
 
-If pitest did not split tests in this way then it would have to keep executing tests even if the first test method in a class failed. Whether they extra complexity is worth it depends very much on how the tests are written and how fast each test runs.
+If pitest did not split tests in this way then it would have to keep executing tests even if the first test method in a class failed. Whether the worthiness of such extra complexity strongly depends on how the tests are written and how fast each test runs.
```
maxgabut: `Whether the worthiness of such extra complexity strongly depends on` ? I'm not a native speaker, so I can be wrong, but it sounds strange to me. Maybe drop the `Whether` entirely ?

errandir: I think you are totally right. Maybe this variation will sound better:
```suggestion
If pitest didn't split tests this way, it would have to execute all tests in a class even if the first of them failed. But the worthiness of such extra complexity strongly depends on how tests are written and how fast each test runs.
```

PR #608, opened 2019-05-17
Fixes for issue 597

jon-bell:
The coverage changes I introduced in 1.4.7 will break if more that `Short.MAX_VALUE` classes are loaded - this patch resolves that limitation.

This patch also adds a defensive `Thread.interrupted()` check before copying class data in the `JavassistCoverageInterceptor`. The NIO utilities will throw a `ClosedByInterruptException` if the thread was interrupted (and didn't have its flag cleared) before these methods are called. Poorly-behaved tests might interrupt a thread, not clear the interrupt flag, then trigger class loading, which would then result in this exception.


hcoles:
@jon-bell Thanks for looking at this, will try and get the rest of the coverage changes merged in soon.

PR #614, opened 2019-05-27
Enhance error message on not enabled test plugin

szpak:
I ran into that in https://github.com/szpak/gradle-pitest-plugin/issues/129 - when the testng plugin was not enabled.
If I understand correctly that error message means that no JUnit 4.6+ is on a classpath and no test plugin is enabled for other test tools (assuming they are on a classpath).

Feel free to adjust the message to your preferences.


hcoles:
Thanks!

PR #640, opened 2019-08-26
Allow to fail on one survivor from commandline like it is also for Maven Mojo

Vampire:


Vampire:
Closed, as the exact same change was done in #858 1.5 years later, so this is obsolete.

PR #653, opened 2019-08-27
Make ManifestUtilsTest also work correctly on Windows

Vampire:


Vampire:
The fix in the other PR actually is incomplete. Here it does not work (I tried it exactly like that first).

PR #685, opened 2019-10-01
Print mutation details on inlined code detection warning

Vampire:


Vampire:
I updated this PR to latest `master`, maybe you have time for it now. :-)

Vampire:
I don't fully remember as it is 3 years ago.
But back then I probably greatly missed this information and needed it to properly process without much debugging.
Also, this additional info is on a warning.
Warnings are hopefully not happen that often and if they happen, the average developer should try to get rid of them and welcome to have the necessary information for that. :-)

Ok, that was a lie, the average developer ignores warnings, but don't destroy my dreams. :-D

Vampire:
If you are concerned about the noise, how about adding the details only in `VERBOSE` and `VERBOSE_NO_SPINNER`?

PR #687, opened 2019-10-01
Split Spock specifications into test units

Vampire:


Vampire:
Closed as Spock 1.x is not JUnit 4 based anymore, but a JUnit 5 platform engine, so this is obsolete unless 1.3 should be supported.

PR #700, opened 2019-11-05
Allow custom locations for Test Classes. 

denyska:
Default behavior (i.e. `not(DefaultDependencyPathPredicate())`) assumes that compiled test classes are ALWAYS flatten and are not in JAR/ZIP archives. 

However, BUCK [java_test](https://buck.build/rule/java_test.html) target puts test classes into JARs. 

I guess BAZEL does the same.


denyska:
@hcoles any questions on this one?

Issue #705, opened 2019-11-27
Path name of the subject program


ali-ghanbari:
When the subject program is located in a directory whose name contains a plus (i.e. the character '+'), the latest version of PIT fails to generate any mutations. As for reproducing an experiment that reveals the bug/missing feature, please consider running the following commands in a Unix machine:
```sh
mkdir a+b
cd a+b
git clone git clone https://github.com/apache/commons-cli.git
cd commons-cli
# assuming JAVA_HOME is set to JDK 1.8 base directory
# configure pom.xml file so that it has PIT plugin with 1 thread of execution and all mutators activated. 
mvn clean test
# all tests should pass for 64900674a1d10f51ca9ef61423d6695be17d8c33
mvn org.pitest:pitest-maven:mutationCoverage
```
Here is how plugin config in `pom.xml` looks like:
```xml
<plugin>
	<groupId>org.pitest</groupId>
	<artifactId>pitest-maven</artifactId>
	<version>1.4.10</version>
	<configuration>
	  <threads>1</threads>
	  <mutations>
	    <mutation>ALL</mutation>
	  </mutations>
	</configuration>
</plugin>
```
After invoking PIT, I see an error like this:
```
12:28:07 PM PIT >> INFO : Calculated coverage in 1 seconds.
12:28:08 PM PIT >> INFO : Created  0 mutation test units
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  1.893 s
[INFO] Finished at: 2019-11-27T12:28:08-06:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.pitest:pitest-maven:1.4.10:mutationCoverage (default-cli) on project commons-cli: Execution default-cli of goal org.pitest:pitest-maven:1.4.10:mutationCoverage failed: No mutations found. This probably means there is an issue with either the supplied classpath or filters.
[ERROR] See http://pitest.org for more details.
[ERROR] -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/PluginExecutionException
```
[...]

PR #706, opened 2019-11-27
solved issue #705: '+' in path name is now allowed

ali-ghanbari:
I think we can work around issue #705 by replacing every "+" with "\\+" in the regular expressions used in the class org.pitest.util.Glob.

### DIFF
--- a/pitest/src/main/java/org/pitest/util/Glob.java
+++ b/pitest/src/main/java/org/pitest/util/Glob.java
@@ -26,12 +26,14 @@ public class Glob implements Predicate<String> {
   private final Pattern regex;
 
   public Glob(final String glob) {
+    String rectifiedGlob;
     if (glob.startsWith("~")) {
-      this.regex = Pattern.compile(glob.substring(1));
+      rectifiedGlob = glob.substring(1);
     } else {
-      this.regex = Pattern.compile(convertGlobToRegex(glob)
-      );
+      rectifiedGlob = convertGlobToRegex(glob);
     }
+    rectifiedGlob = rectifiedGlob.replace("+", "\\+");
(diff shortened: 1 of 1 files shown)


hcoles:
Is this confirmed to resolve the issue? I would only expect globs to be matched package names, rather than the containing folder.

ali-ghanbari:
I did several tests and it works.

Printing glob values revealed that path names are also passed as glob values, and that causes the aforementioned issue. I guess, `org.pitest.mutationtest.config.ReportOptions` is where globs are used for non-package related matching.

StefanPenndorf:
I think there are lots more special characters to mention. Same applies for (){}[] under Windows and Linux. Backslash \ is unusual but allowed under linux as well.

@ali-ghanbari Can you add some test cases for the paths you saw being passed to Glob? Maybe that helps adding more cases and finding an appropriate solution.

ali-ghanbari:
@StefanPenndorf, thank you for mentioning those other characters! I checked [Pattern](https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html), and found more characters that might help you debug the program.

Some of the characters, like [] and () are tricky as when only the right one occurs Glob works correctly. I tried to exercise Glob with different combinations and I created interesting test cases. Most of them fail, hence localize the bug, while some of them pass. I think we might have all of them as more test cases won't hurt.

Lastly, I used more or less the same strings that were passed to Glob. This made the test cases a bit complicated, but I think we can always shorten the strings appropriately if needed.

hcoles:
@ali-ghanbari thanks for this, sorry its been sat around for so long. Your fix has been applied in #1031 and will be merged in shortly.

Issue #723, opened 2020-01-20
Update vulnerable libraries in pitest-maven


jimbethancourt:
The OWASP Dependency Check has identified several transitive dependencies have critical or high severity:

- commons-collections-3.2.jar (pkg:maven/commons-collections/commons-collections@3.2, cpe:2.3:a:apache:commons_collections:3.2:*:*:*:*:*:*:*) : CVE-2015-6420, CVE-2017-15708, Remote code execution
- commons-beanutils-1.7.0.jar (pkg:maven/commons-beanutils/commons-beanutils@1.7.0, cpe:2.3:a:apache:commons_beanutils:1.7.0:*:*:*:*:*:*:*) : CVE-2014-0114, CVE-2019-10086
- plexus-utils-1.5.8.jar (pkg:maven/org.codehaus.plexus/plexus-utils@1.5.8, cpe:2.3:a:plexus-utils_project:plexus-utils:1.5.8:*:*:*:*:*:*:*) : CVE-2017-1000487, Directory traversal in org.codehaus.plexus.util.Expand, Possible XML Injection
- groovy-all-1.7.6.jar (pkg:maven/org.codehaus.groovy/groovy-all@1.7.6) : CVE-2015-3253, CVE-2016-6814

I have updated with suitable workarounds and tested report generation locally and will create a PR soon.

PR #724, opened 2020-01-20
#723 Upgraded versions of vulnerable transitive dependencies in pitest-maven

jimbethancourt:
Upgraded versions of vulnerable transitive dependencies in pitest-maven and adding the OWASP Dependency Check plugin to the pitest-parent POM

Upgraded versions of vulnerable transitive dependencies, placing them immediately before the dependencies where they are used to ensure that Maven excludes the transitive dependencies:
- commons-collections, commons-beanutils, and plexus-utils are all used by maven-reporting-impl
- groovy-all is a transitive dependency of maven-scm-providers-standard.  It is used by maven-scm-provider-integrity

Added the OWASP Dependency Check plugin to the pitest-parent POM to allow it to be run in any module.

### DIFF
--- a/pitest-maven/pom.xml
+++ b/pitest-maven/pom.xml
@@ -102,6 +102,8 @@
 			<artifactId>maven-reporting-api</artifactId>
 			<version>${maven.version}</version>
 		</dependency>
+		<!-- commons-collections, commons-beanutils, and plexus-utils specified
+		 due to vulnerable versions used by maven-reporting-impl:2.0.4.3-->
 		<dependency>
 			<groupId>commons-collections</groupId>
 			<artifactId>commons-collections</artifactId>
@@ -137,6 +139,8 @@
 			<artifactId>maven-scm-manager-plexus</artifactId>
 			<version>${scm.version}</version>
 		</dependency>
+		<!-- Specifying groovy-all because maven-scm-providers-standard:1.9.4
+		 relies on a vulnerable version of groovy-all-->
 		<dependency>
 			<groupId>org.codehaus.groovy</groupId>
 			<artifactId>groovy-all</artifactId>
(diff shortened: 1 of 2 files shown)


szpak:
> Upgraded versions of vulnerable transitive dependencies, placing them immediately before the dependencies where they are used to ensure that Maven excludes the transitive dependencies

Have you tried to upgrade a dependency(ies) that includes those vulnerable transitive dependencies?

jimbethancourt:
Yes - unfortunately the vulnerabilities exist in when the dependencies that include vulnerable transitive dependencies are upgraded, and introduce more vulnerable libraries as well :-(

szpak:
That's sad.

Could you link here issues in those project(s) to allow us to track it and remove those implicit dependencies once it's (eventually) fixed upstream?

hcoles:
@jimbethancourt Thanks for this.

The owasp plugin looks to be bound to an early maven phase and runs by default, significantly slowing  the build and possibly breaking it if a new vulnerability is disclosed.before changes are committed.

If you can bind it to a profile so it is invoked explicitly, and add comments to the pom noting why the dependencies have been added, I'll merge it in.

jimbethancourt:
> That's sad.
> 
> Could you link here issues in those project(s) to allow us to track it and remove those implicit dependencies once it's (eventually) fixed upstream?

@szpak I will see what I can do.

jimbethancourt:
> @jimbethancourt Thanks for this.
> 
> The owasp plugin looks to be bound to an early maven phase and runs by default, significantly slowing the build and possibly breaking it if a new vulnerability is disclosed.before changes are committed.
> 
> If you can bind it to a profile so it is invoked explicitly, and add comments to the pom noting why the dependencies have been added, I'll merge it in.

Sure!  I will take care of that this weekend.

jimbethancourt:
Hi @hcoles - I've updated the PR.  I'm proud to say that it was done at the Houston Java Users Group meeting where I presented on Mutation Testing and PIT. :-)

hcoles:
thanks!

PR #733, opened 2020-02-12
replace Preconditions.checkNotNull with Objects.requireNonNull

AlexElin:
replace Preconditions.checkNotNull with Objects.requireNonNull from Java Standard API


hcoles:
Thanks!

PR #736, opened 2020-02-14
replace StringUtil.join with java.lang.String.join

AlexElin:
replace StringUtil.join with java.lang.String.join


hcoles:
Thanks for all these PRs. A lot of the code was originally written against java 5, and it shows.

PR #738, opened 2020-02-22
Replace Prelude with native Java 8

simon04:


pitest/src/main/java/org/pitest/coverage/execute/CoverageMinion.java
```diff
@@ -116,8 +115,8 @@ public static void main(final String[] args) {
 
   private static void enablePowerMockSupport() {
     // Bwahahahahahahaha
-    HotSwapAgent.addTransformer(new BendJavassistToMyWillTransformer(Prelude
-        .or(new Glob("javassist/*")),
+    HotSwapAgent.addTransformer(new BendJavassistToMyWillTransformer(
+        new Glob("javassist/*"),
```
StefanPenndorf, line 119: new Glob vs. toGlobPredicate() -- would rather use one approach either the factory method or the constructor but we shouldn't mix both. Maybe we should go with the factory method. What do you think?

pitest/src/main/java/org/pitest/coverage/execute/CoverageOptions.java
```diff
@@ -47,9 +47,11 @@ public CoverageOptions(final Collection<String> include, final Collection<String
   }
 
   public Predicate<String> getFilter() {
-    return Prelude.and(Prelude.or(Glob.toGlobPredicates(this.include)),
-        Prelude.not(Prelude.or(Glob.toGlobPredicates(this.exclude))),
-        Prelude.not(commonClasses()));
+    return Stream.of(
+        Glob.toGlobPredicate(this.include),
+        Glob.toGlobPredicate(this.exclude).negate(),
+        commonClasses().negate()
```
StefanPenndorf, line 53: Instead of Streaming what about

```
  return Glob.toGlobPredicate(this.include)
  .and(Glob.toGlobPredicate(this.exclude).negate())
  .and(commonClasses().negate())
```

pitest/src/main/java/org/pitest/mutationtest/engine/gregor/GregorMutater.java
```diff
@@ -118,8 +117,12 @@ public Mutant getMutation(final MutationIdentifier id) {
   }
 
   private Predicate<MethodInfo> filterMethods() {
-    return and(this.filter, filterSyntheticMethods(),
-        isGeneratedEnumMethod().negate(), isGroovyClass().negate());
+    return Stream.of(
```
StefanPenndorf, line 120: Hmm stream or method chain. Seems to be a matter of taste?

pitest-entry/src/main/java/org/pitest/mutationtest/config/ReportOptions.java
```diff
@@ -78,7 +75,7 @@
   private Collection<String>             excludedClasses                = Collections
       .emptyList();
 
-  private Collection<Predicate<String>>  excludedTestClasses            = Collections
+  private Collection<String>             excludedTestClasses            = Collections
```
hcoles, line 78: This is quite a large PR. but largely easy to merge as it is just replacing legacy collection manipulation with more modern ones.

A signature/type change is different. I'm not sure if switching to a Collection<String> is desirable, or if it has wider consequences. I'd prefer not to have to think about that before merging this PR, could you revert this?

simon04:
Sigh… It took me at least half an hour to resolve the merge conflict from the recent changes. It found it very hard to amend this PR after two months of inactivity.

> A signature/type change is different. I'm not sure if switching to a Collection is desirable, or if it has wider consequences. I'd prefer not to have to think about that before merging this PR, could you revert this?
I tried to undo this part. However, after 20min of trying, I gave up. Either take it as it is or forget this PR, sorry.

pitest/src/main/java/org/pitest/coverage/execute/CoverageOptions.java
simon04: :+1:

pitest-entry/src/main/java/org/pitest/mutationtest/build/intercept/javafeatures/InlinedFinallyBlockFilter.java
```diff
@@ -97,7 +95,7 @@ private void checkForInlinedCode(final Collection<MutationDetails> combined,
     // to the base one (is this not implied by there being only 1 mutation in
     // the handler ????)
     final List<Integer> ids = map(each.getValue(), mutationToBlock());
-    if (ids.stream().filter(not(isEqual(firstBlock))).findFirst().isPresent()) {
+    if (ids.stream().filter(Predicate.<Integer>isEqual(firstBlock).negate()).findFirst().isPresent()) {
```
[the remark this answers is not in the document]
simon04: :+1:

pitest-entry/src/main/java/org/pitest/mutationtest/build/DefaultTestPrioritiser.java
```diff
@@ -69,8 +68,7 @@ public DefaultTestPrioritiser(CoverageDatabase coverage) {
 
   private List<TestInfo> prioritizeTests(ClassName clazz,
       Collection<TestInfo> testsForMutant) {
-    final List<TestInfo> sortedTis = FCollection.map(testsForMutant,
-        Prelude.id(TestInfo.class));
+    final List<TestInfo> sortedTis = FCollection.map(testsForMutant, x -> x);
```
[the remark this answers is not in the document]
simon04: Already done in 0b8b5826cb2d760ede8f23a3adc9f38e9e05daa4

pitest-aggregator/src/main/java/org/pitest/aggregate/CodeSourceAggregator.java
```diff
@@ -49,7 +48,9 @@ private ProjectClassPaths createProjectClassPaths() {
         classes.addAll(FCollection.map(dcRoot.classNames(), toPredicate()));
       }
     }
-    return Prelude.or(FCollection.map(classes, Glob.toGlobPredicate()));
+    return classes.stream()
+        .map(Glob.toGlobPredicate())
```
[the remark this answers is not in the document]
simon04: :+1:

pitest/src/main/java/org/pitest/mutationtest/engine/gregor/GregorMutater.java
simon04: I changed it to method chain.

pitest-entry/src/main/java/org/pitest/mutationtest/config/ReportOptions.java
StefanPenndorf: I don't have a lot of experience with Github reviews, but if I'm not missing anything, that hasn't been done yet, has it?

pitest/src/test/java/org/pitest/functional/FCollectionTest.java
```diff
@@ -67,12 +66,12 @@ public void shouldApplyForEachToAllItems() {
 
   @Test
   public void shouldApplyMapToAllItems() {
-    assertEquals(this.is, FCollection.map(this.is, Prelude.id()));
+    assertEquals(this.is, FCollection.map(this.is, x -> x));
```
StefanPenndorf, line 69: There is ``java.util.function.Function.identity()`` which does exactly the same. Would have replaced Prelude.id() by this built-in function but ``x -> x`` does exactly the same. Maybe a matter of taste.... Don't feel it's worth insisting on changing that. Code is readable anyway.

StefanPenndorf (changes requested):
Hi @simon04 thank you for being so patient with us. Could you resolve @hcoles's comment in ``ReportOptions``? Or am I missing something?

pitest-entry/src/main/java/org/pitest/mutationtest/config/ReportOptions.java
simon04: No, please see my https://github.com/hcoles/pitest/pull/738#issuecomment-622063578. I'm not going to spend any more time on this PR.

PR #753, opened 2020-04-07
migrate to mockito 3

AlexElin:
migrate to mockito 3: update versions, fix tests and deprecated API usage.


pitest-entry/src/test/java/org/pitest/mutationtest/incremental/IncrementalAnalyserTest.java
```diff
@@ -59,7 +60,7 @@ public void shouldStartPreviousSurvivedMutationsAtAStatusOfNotStartedWhenCoverag
     setHistoryForAllMutationsTo(DetectionStatus.SURVIVED);
     when(
         this.history.hasCoverageChanged(any(ClassName.class),
-            any(BigInteger.class))).thenReturn(true);
+            isNull())).thenReturn(true);
```
StefanPenndorf, line 63: This change doesn't feel right. DefaultCodeHistory.hasCoverageChanged will throw a NullPointerException when called with NULL. I think we should put more effort into mocking in order to test correct interaction with collaborators:
```java
    when(this.coverage.getCoverageIdForClass(ClassName.fromString("clazz")))
            .thenReturn(BigInteger.valueOf(42L));
    when(this.history.hasCoverageChanged(ClassName.fromString("clazz"), BigInteger.valueOf(42L)))
            .thenReturn(true);
```
(feel free to extract constants and beautify the code above)

pitest-entry/src/test/java/org/pitest/mutationtest/config/CompoundListenerFactoryTest.java
```diff
@@ -52,12 +50,8 @@ public void setUp() {
   public void shouldCreateACombinedListenerForAllChildFactories() {
     final MutationResultListener listenerOne = mock(MutationResultListener.class);
     final MutationResultListener listenerTwo = mock(MutationResultListener.class);
-    when(
-        this.firstChild.getListener(any(Properties.class),
-            any(ListenerArguments.class))).thenReturn(listenerOne);
-    when(
-        this.secondChild.getListener(any(Properties.class),
-            any(ListenerArguments.class))).thenReturn(listenerTwo);
+    when(this.firstChild.getListener(isNull(), isNull())).thenReturn(listenerOne);
```
StefanPenndorf, line 53: @AlexElin would you mind improving this test? You're completely right in changing the expectation from any() to isNull() but this reveals that the test is ... incomplete at least...

I think we should add another test in order to check that the non-null arguments provided to  testee will be forwarded to first and second child. Do you agree?

pitest-maven/src/test/java/org/pitest/maven/MojoToReportOptionsConverterTest.java
```diff
@@ -376,7 +377,7 @@ public void testParsesExcludedClasspathElements()
   public void testParsesSurefireConfigWhenFlagSet() {
     parseConfig("<parseSurefireConfig>true</parseSurefireConfig>");
     verify(this.surefireConverter).update(any(ReportOptions.class),
-        any(Xpp3Dom.class));
+        isNull());
```
StefanPenndorf, line 380: From looking at the code of MojoToReportOptionsConverter and maven classes I doubt this will ever be null. But mocking this out will be cumbersome. Maybe that's not worth the effort.

pitest/src/test/java/org/pitest/coverage/execute/DependencyFilterTest.java
```diff
@@ -64,13 +65,13 @@ public void shouldNotPerformAnalysisWhenDependencyDistanceIsLessThan0()
...
-            .getDescription().getFirstTestClass()), any(Predicate.class)))
-            .thenReturn(Arrays.asList("foo"));
+        this.extractor.extractCallDependenciesForPackages(
+            eq(this.aTestUnit.getDescription().getFirstTestClass()),
+            isNull())).thenReturn(Arrays.asList("foo"));
     when(
         this.extractor.extractCallDependenciesForPackages(
             eq(this.anotherTestUnit.getDescription().getFirstTestClass()),
-            any(Predicate.class))).thenReturn(Collections.<String> emptyList());
+            isNull())).thenReturn(Collections.<String> emptyList());
```
StefanPenndorf, line 74: This null should never happen because it will lead to a NPE in DependencyExtractor.

I think it would be better to mock the Predicate<String> here as well (used as second constructor paramter).

pitest-entry/src/test/java/org/pitest/mutationtest/config/CompoundListenerFactoryTest.java
hcoles: `isNull` is rarely the right matcher. If null is being passed around it is most likely an artefact of the tests, rather than part of any contract.

Generally, stubs should be as general and accepting as possible (so any(Class) rather than restricting to null values). 

Verifications should be as precise as necessary (so rarely use any, unlikely to us isNull).

pitest-entry/src/test/java/org/pitest/mutationtest/incremental/IncrementalAnalyserTest.java
hcoles: (also welcome back @StefanPenndorf, not seen you for a long while).

pitest/src/test/java/org/pitest/coverage/execute/DependencyFilterTest.java
AlexElin: I've added always true predicate into `DependencyFilter` instead of `null`

pitest-entry/src/test/java/org/pitest/mutationtest/incremental/IncrementalAnalyserTest.java
AlexElin: fixed.

pitest-entry/src/test/java/org/pitest/mutationtest/config/CompoundListenerFactoryTest.java
AlexElin: changed.
Passed non-null arguments to `this.testee.getListener(...)`

pitest-maven/src/test/java/org/pitest/maven/MojoToReportOptionsConverterTest.java
AlexElin: The reason of `null `is that `surifire` (MojoToReportOptionsConverterTest:54) has no configuration

AlexElin:
@StefanPenndorf, @hcoles I've fixed all your comments.
Could you re-review changes and merge it (if there will be no new comments)?

pitest-maven/src/test/java/org/pitest/maven/PitMojoTest.java
```diff
@@ -37,7 +37,7 @@ public void testRunsAMutationReportWhenMutationCoverageGoalTrigered()
     build.setOutputDirectory("foo");
     this.testee.getProject().setBuild(build);
     this.testee.execute();
-    verify(this.executionStrategy).execute(any(File.class),
```
StefanPenndorf, line 40: In line with Henry's comment on `CompoundListenerFactoryTest`, we should stick to ```any(File.class)``` and, if in doubt, use a real file object or mock. But those tests do not aim for testing the base dir parameter gets correctly handed over to the strategy. Thus I would prefer ```any(File.class)``` everywhere in this test class.
This applies to all 4 changes you've made.

Maybe we could add another test case that checks the usual case doesn't get broken. Something like

```java
  @Mock
  private MavenProject executionProject;

  @Override
  public void setUp() throws Exception {
    super.setUp();

    when(this.project.getExecutionProject()).thenReturn(executionProject);
    when(this.executionProject.getBasedir()).thenReturn(BASEDIR);
  }

  public void testUsesExecutionProjectBaseDir() throws Exception {
    this.testee = createPITMojo(createPomWithConfiguration(""));
    this.testee.execute();

    verify(this.executionStrategy).execute(eq(BASEDIR),
            any(ReportOptions.class), any(PluginServices.class), anyMap());
  }
```

StefanPenndorf (changes requested):

pitest-maven/src/test/java/org/pitest/maven/PitMojoTest.java
AlexElin: made changes.
I preserved `any(File.class)`.
To avoid `baseDir `as `null` I mocked `executionProject`

AlexElin:
I've updated mockito to the latest version 3.6.28
Additionaly I replaced deprecated `MockitoAnnotations.initMocks` with `MockitoAnnotations.openMocks`

StefanPenndorf (approved):
@AlexElin  Thank you for your patience!

@hcoles LGMT -- If you like the change too, it's worth merging it soon, otherwise there will likely be merge conflicts to resolve sooner or later.

AlexElin:
@hcoles could you merge the changes? if you have no objections, of course

hcoles:
Hi @AlexElin, thanks for the MR and sorry for the long delay. Fortunately I finally have a bit of time available for pitest.

PR #757, opened 2020-04-17
use java 8 features: List.sort, Comparator.comparing and others

AlexElin:
use java 8 features: List.sort, Comparator.comparing and others

### DIFF
--- a/pitest-entry/src/main/java/org/pitest/bytecode/analysis/MethodTree.java
+++ b/pitest-entry/src/main/java/org/pitest/bytecode/analysis/MethodTree.java
@@ -74,9 +74,8 @@ public class MethodTree {
 
   private List<AbstractInsnNode> createInstructionList() {
     final List<AbstractInsnNode> list = new LinkedList<>();
-    final ListIterator<AbstractInsnNode> it = this.rawNode.instructions.iterator();
-    while (it.hasNext()) {
-        list.add(it.next());
+    for (AbstractInsnNode abstractInsnNode : this.rawNode.instructions) {
+      list.add(abstractInsnNode);
     }
     this.lazyInstructions = new ArrayList<>(list);
     return this.lazyInstructions;
--- a/pitest-entry/src/main/java/org/pitest/bytecode/analysis/MethodTree.java
+++ b/pitest-entry/src/main/java/org/pitest/bytecode/analysis/MethodTree.java
@@ -1,7 +1,6 @@
 package org.pitest.bytecode.analysis;
 
 import java.util.ArrayList;
-import java.util.LinkedList;
 import java.util.List;
 
 import org.objectweb.asm.Opcodes;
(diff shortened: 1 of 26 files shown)


pitest-entry/src/main/java/org/pitest/bytecode/analysis/MethodTree.java
```diff
@@ -74,9 +73,8 @@ public boolean isSynthetic() {
 
   private List<AbstractInsnNode> createInstructionList() {
     final List<AbstractInsnNode> list = new LinkedList<>();
-    final ListIterator<AbstractInsnNode> it = this.rawNode.instructions.iterator();
-    while (it.hasNext()) {
-        list.add(it.next());
+    for (AbstractInsnNode abstractInsnNode : this.rawNode.instructions) {
+      list.add(abstractInsnNode);
     }
     this.lazyInstructions = new ArrayList<>(list);
```
StefanPenndorf, line 79: I'm wondering why there is a double copy of the list. First when iterating over this.rawNode.instructions those instructions will be "copied"  into list. Afterwards list will be copied into new ArraList<>(list) . I don't think the second copy is necessary. We should use ArrayList from the beginning. Is the size/number of instructions known? Would make sense to initialize the ArrayList with the number of items before copying from instructions.

pitest-entry/src/main/java/org/pitest/coverage/CoverageData.java
```diff
@@ -252,13 +251,7 @@ private TestInfo createTestInfo(final Description description,
   }
 
   private int getNumberOfCoveredLines(final ClassName clazz) {
```
StefanPenndorf, line 253: The name of the method and the implementation do not fit, what do you think (is not your fault but as I review the code that came to my mind) the method returns theNumberOfCoveredTests(...) not the number of covered lines. Need to check usages of the method later.

pitest/src/test/java/org/pitest/functional/FCollectionTest.java
```diff
@@ -60,9 +60,9 @@ public void shouldApplyForEachToAllItems() {
     final List<Integer> actual = new ArrayList<>();
     final Consumer<Integer> e = a -> actual.add(a);
 
-    FCollection.forEach(this.is, e);
+      this.is.forEach(e);
 
-    assertEquals(this.is, actual);
+      assertEquals(this.is, actual);
```
StefanPenndorf, line 65: The additional indention is strange ... but doesn't harm anybody ;-) Maybe the whole method is not correctly indented? I didn't check.

AlexElin: fixed

pitest-entry/src/main/java/org/pitest/bytecode/analysis/MethodTree.java
AlexElin: removed double copy.

pitest-entry/src/main/java/org/pitest/coverage/CoverageData.java
AlexElin: I think `getTestsForClassName` has wrong name.
Tried to resolve this problem in latest commit.
Please, take a look

StefanPenndorf: LGTM

PR #758, opened 2020-04-18
Find occupied test packages if targetTests not provided, the same way it is done for targetClasses

nicerloop:


hcoles:
Thanks @nicerloop, from a scan over this looks correct. If you can add a integration test that confirms it works to the maven-verification module I'll merge it in.

PR #765, opened 2020-05-06
Allowed more tests to be considered killing a mutant when using historic data (fixes #763)

### LINKED ISSUE #763: Increment Analysis Issue
Increment analysis is not working correctly.

				<plugin>
				<groupId>org.pitest</groupId>
				<artifactId>pitest-maven</artifactId>
				<version>1.4.11</version>
				<configuration>
					<testPlugin>testng</testPlugin>
					<verbose>true</verbose>
					<features>
						<value>+EXPORT</value>
					</features>
					<targetClasses>
						<param>com.dt.appservice.*</param>
					</targetClasses>
					<historyInputFile>${project.basedir}\\test-output\\PitestHistory\\history.txt</historyInputFile>
					<historyOutputFile>${project.basedir}\\test-output\\PitestHistory\\history.txt</historyOutputFile>
					<targetTests>
						<param>com.abc.dtc.service.*</param>
					</targetTests>
					
						</configuration>
					<plugin>

It is creating history file  but not referring  for next run.. When I am running same set of files without  any modifications, It is taking same time. No performance improvement.

could you help me on this?

StefanPenndorf:
fixes #763

Incremental analysis honors all killing tests now instead of limiting the scope of analysis to the first killing test.
This is often identical if there is no matrix mutation in place but TestNG plugin calculates test units in terms of test classes thus executes all test methods of a test class and records historic data
with method level granularity and class level granularity. We need to take all killing tests into account to match history data and executable test units in the analyzer. This will also help if different test classes cover a certain production class and one of those test classes is deleted or renamed.

Improved logging to help spotting analyzer misbehaviour. Although pitest is already verbose I added another info-level statement.

I've also refactored the test and added more test cases to prove the correctness of my changes.

### DIFF
--- a/pitest-entry/src/test/java/org/pitest/mutationtest/incremental/IncrementalAnalyserTest.java
+++ b/pitest-entry/src/test/java/org/pitest/mutationtest/incremental/IncrementalAnalyserTest.java
@@ -1,15 +1,35 @@
 package org.pitest.mutationtest.incremental;
 
-import static org.junit.Assert.assertEquals;
+import static java.util.Arrays.asList;
+import static java.util.Collections.emptyList;
+import static java.util.Collections.singletonList;
+import static org.hamcrest.Matchers.allOf;
+import static org.hamcrest.Matchers.contains;
+import static org.hamcrest.Matchers.hasItem;
+import static org.hamcrest.Matchers.hasItems;
+import static org.junit.Assert.assertThat;
 import static org.mockito.Matchers.any;
 import static org.mockito.Mockito.when;
+import static org.pitest.mutationtest.DetectionStatus.KILLED;
+import static org.pitest.mutationtest.DetectionStatus.NOT_STARTED;
+import static org.pitest.mutationtest.DetectionStatus.SURVIVED;
+import static org.pitest.mutationtest.DetectionStatus.TIMED_OUT;
 import static org.pitest.mutationtest.LocationMother.aLocation;
 import static org.pitest.mutationtest.LocationMother.aMutationId;
 
 import java.math.BigInteger;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Optional;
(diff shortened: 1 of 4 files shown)


pitest-entry/src/test/java/org/pitest/mutationtest/incremental/IncrementalAnalyserTest.java
```diff
@@ -34,23 +54,41 @@
...
   }
 
+  private void configureLogCatcher() {
+    logCatcher = new LogCatcher();
+    final Logger logger = Log.getLogger();
+
+    for (Handler handler : logger.getHandlers()) {
+      logger.removeHandler(handler);
+    }
+    logger.addHandler(logCatcher);
```
hcoles, line 66: @StefanPenndorf Are we introducing a memory leak into the test suite here? If nothing removes the LogCatcher when the last runs then presumably it continues to accumulate log entries until the JVM exits?

StefanPenndorf: Yes, my first commit had this issue: It introduced a memory leak as well as changing the behaviour for subsequent tests. Thus I decided not to disable default handlers by removing them and to remove the log catcher in a tear down method (annotated with ```@After```)

I guess you found the fix because you merged my change in anyway.

hcoles: Saw your change come in so merged it - hopefully not prematurely?

Tracked down a leak exactly like this one in a commercial test suite years ago, so knew to look for it when testing logging.

StefanPenndorf: Everything well, pull request was complete with two commits. Would have left a note if I knew there was something wrong/missing.

Issue #537, opened 2018-11-29
PitError: Coverage generation minion exited abnormally!


michielgkalkman:
Please have a look at https://github.com/michielgkalkman/PmdPlus/commit/fd84044f487517307c5b3772841d710f4179da3e

When running mvn -e clean install org.pitest:pitest-maven:mutationCoverage

I get a PitError: Caused by: org.pitest.util.PitError: Coverage generation minion exited abnormall
y!

See the entire stacktrace below.

BTW: maybe it is an idea to add the location where the issue needs to be reported, because I wondered for while whether it was a vm issue or not.

Thanks and regards,
Michiel

Caused by: org.pitest.util.PitError: Coverage generation minion exited abnormall
y!

Please copy and paste the information and the complete stacktrace below when rep
orting an issue
VM : Java HotSpot(TM) 64-Bit Server VM
Vendor : Oracle Corporation
Version : 25.191-b12
Uptime : 24878
Input ->
 1 : -Dclassworlds.conf=C:\j3\apache-maven\apache-maven-3.5.4\bin\..\bin\m2.conf

 2 : -Dmaven.home=C:\j3\apache-maven\apache-maven-3.5.4\bin\..
 3 : -Dlibrary.jansi.path=C:\j3\apache-maven\apache-maven-3.5.4\bin\..\lib\jansi
-native
 4 : -Dmaven.multiModuleProjectDirectory=C:\Users\nl02032\projects\PmdPlus
BootClassPathSupported : true

Please copy and paste the information and the complete stacktrace below when rep
orting an issue
VM : Java HotSpot(TM) 64-Bit Server VM
Vendor : Oracle Corporation
Version : 25.191-b12
Uptime : 24879
Input ->
 1 : -Dclassworlds.conf=C:\j3\apache-maven\apache-maven-3.5.4\bin\..\bin\m2.conf

 2 : -Dmaven.home=C:\j3\apache-maven\apache-maven-3.5.4\bin\..
 3 : -Dlibrary.jansi.path=C:\j3\apache-maven\apache-maven-3.5.4\bin\..\lib\jansi
-native
 4 : -Dmaven.multiModuleProjectDirectory=C:\Users\nl02032\projects\PmdPlus
BootClassPathSupported : true

    at org.pitest.util.Unchecked.translateCheckedException (Unchecked.java:20)
    at org.pitest.coverage.execute.DefaultCoverageGenerator.calculateCoverage (D
efaultCoverageGenerator.java:103)
[...]

PR #769, opened 2020-06-01
Fix for issue 537

michielgkalkman:
Hi all,

This is my fix for https://github.com/hcoles/pitest/issues/537.

Appearently, it is possible to get a Description from a org.junit.runner.Runner with a non-existent classname (or any name actually). This resulted in a ClassNotFoundException when there were no child descriptions, or it resulted in no TestUnits if there were child Descriptions.

I think I fixed those issues. At least the other unit tests still work and my issue is solved :)

Thanks and regards,
Michiel

### DIFF
--- a/pitest/src/main/java/org/pitest/junit/JUnitCustomRunnerTestUnitFinder.java
+++ b/pitest/src/main/java/org/pitest/junit/JUnitCustomRunnerTestUnitFinder.java
@@ -46,6 +46,7 @@ import org.pitest.testapi.TestGroupConfig;
 import org.pitest.testapi.TestUnit;
 import org.pitest.testapi.TestUnitFinder;
 import org.pitest.util.IsolationUtils;
+import org.pitest.util.PitError;
 
 public class JUnitCustomRunnerTestUnitFinder implements TestUnitFinder {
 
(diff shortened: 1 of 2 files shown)


hcoles:
Thanks @michielgkalkman, did you try the simpler solution of adding the PMD runner to the blacklist of runners that should not be split? Did that work?

michielgkalkman:
Somehow I got the impression that your suggestion was not to be the solution, but it turns out that blacklisting works!

However, it does have the disadvantages that 1) one explicitly needs to add a fully qualified classname (I would at least prefer a configuration option so you can exclude other test runners as well) and 2) the unit tests I provided still fail.

The solution in this pull request might be useful for other test runners. I intend to check out one of the blacklisted test runners and check why they go wrong and if this pull request somehow helps.

Anyway: there seems to be 3 options:
1) add the PmdTestRunner to the blacklist and ignore the unit tests results
2) Use my solution
3) Use the whitelist solution

I know too little of pitest code to be able to say something useful about option 3. Does it have a big impact? If so, it might seem useful to select option 1 or 2 at first and prepare a future release to use option 3.

WDYT?

michielgkalkman:
I was not able to find a situation where removing an artifact from the blacklist resulted in problems. Does anyone have an example project which goes wrong after removing?

Issue #770, opened 2020-06-03
NPE in enum constructors if hamcrest on classpath


aJanuary:
Since 1.4.7, if I have hamcrest in the classpath I get a `NullPointerException` thrown in the `<init>` method of enums with constructors if I try to load them using `Class.forName` when pitest does the initial test run.

```
package com.example;

import org.junit.Test;

public class NestedClassTest {
	public static enum MyEnum {
		A { }
	}

	@Test
	public void test() throws ClassNotFoundException {
		Class.forName("com.example.NestedClassTest$MyEnum$1");
	}
}
```

```
   [pitest] 12:06:57 PIT >> INFO : MINION : 12:06:57 PIT >> INFO : 1 tests received
   [pitest] 
   [pitest] 12:06:57 PIT >> INFO : MINION : 12:06:57 PIT >> FINE : Running 1 units
   [pitest] 
   [pitest] 12:06:57 PIT >> INFO : MINION : 12:06:57 PIT >> FINE : Gathering coverage for test Description [testClass=com.example.NestedClassTest, name=test(com.example.NestedClassTest)]
   [pitest] 
   [pitest] 12:06:57 PIT >> INFO : MINION : 12:06:57 PIT >> SEVERE : Description [testClass=com.example.NestedClassTest, name=test(com.example.NestedClassTest)]
   [pitest] java.lang.ExceptionInInitializerError
   [pitest] 	at java.lang.Class.forName0(Native Method)
   [pitest] 	at java.lang.Class.forName(Class.java:264)
   [pitest] 	at com.examp
   [pitest] 12:06:57 PIT >> INFO : MINION : le.NestedClassTest.test(Unknown Source)
   [pitest] 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
   [pitest] 	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
   [pitest] 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAc
   [pitest] 12:06:57 PIT >> INFO : MINION : cessorImpl.java:43)
   [pitest] 	at java.lang.reflect.Method.invoke(Method.java:498)
   [pitest] 	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
   [pitest] 	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
[...]

Issue #746, opened 2020-03-12
Tests failing with new coverage collection


LaurentTho3:
Hello,

I am trying to run PIT on the apache commons BCEL project (https://github.com/apache/commons-bcel) version 6.4.0 (tag rel/commons-bcel-6.4.0). When using recent versions of PIT (>= 1.4.7) the build fails, saying that 33 tests did not pass when calculating line coverage although the tests pass when ran without PIT. The tests are returning errors similar to this one:

<details>
  <summary>Example of error returned by tests</summary>

```
Description [testClass=org.apache.bcel.verifier.VerifierTestCase, name=testDefaultMethodValidation(org.apache.bcel.verifier.Verifier
TestCase)]
java.lang.NoClassDefFoundError: Could not initialize class org.apache.bcel.generic.Type
at org.apache.bcel.verifier.statics.Pass2Verifier$CPESSC_Visitor.visitMethod(Pass2Verifier.java:636)
        at org.apache.bcel.classfile.Method.accept(Method.java:108)
        at org.apache.bcel.classfile.DescendingVisitor.visitMethod(DescendingVisitor.java:158)
        at org.apache.bcel.classfile.Method.accept(Method.java:108)
        at org.apache.bcel.classfile.DescendingVisitor.visitJavaClass(DescendingVisitor.java:98)
        at org.apache.bcel.classfile.JavaClass.accept(JavaClass.java:213)
        at org.apache.bcel.classfile.DescendingVisitor.visit(DescendingVisitor.java:84)
        at org.apache.bcel.verifier.statics.Pass2Verifier$CPESSC_Visitor.<init>(Pass2Verifier.java:360)
        at org.apache.bcel.verifier.statics.Pass2Verifier$CPESSC_Visitor.<init>(Pass2Verifier.java:316)
        at org.apache.bcel.verifier.statics.Pass2Verifier.constant_pool_entries_satisfy_static_constraints(Pass2Verifier.java:301)
        at org.apache.bcel.verifier.statics.Pass2Verifier.do_verify(Pass2Verifier.java:160)
        at org.apache.bcel.verifier.PassVerifier.verify(PassVerifier.java:70)
        at org.apache.bcel.verifier.Verifier.doPass2(Verifier.java:75)
        at org.apache.bcel.verifier.VerifierTestCase.testDefaultMethodValidation(VerifierTestCase.java:36)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[...]

PR #771, opened 2020-06-05
Check that PROBE_FIELD_NAME is not null when using it.

LaurentTho3:
Hello,

This is a fix for https://github.com/hcoles/pitest/issues/770, which also seems to solve https://github.com/hcoles/pitest/issues/746.

It looks like in some configurations, it is possible that we try to access the coverage probes field ($$pitCoverageProbes) before it is initialised in the static initialiser (<clinit>). This leads to a NullPointerException.

Here, instead of initialising the field in the static initialiser, we check if it is null every time we access it and initialise it if needed. All the tests are passing.

### DIFF
--- a/pitest/src/main/java/org/pitest/coverage/CoverageClassVisitor.java
+++ b/pitest/src/main/java/org/pitest/coverage/CoverageClassVisitor.java
@@ -39,7 +39,6 @@ public class CoverageClassVisitor extends MethodFilteringAdapter {
   private int       probeCount = 1;
 
   private String    className;
-  private boolean   foundClinit;
 
   public CoverageClassVisitor(final int classId, final ClassWriter writer) {
     super(writer, BridgeMethodFilter.INSTANCE);
@@ -62,10 +61,6 @@ public class CoverageClassVisitor extends MethodFilteringAdapter {
   public MethodVisitor visitMethodIfRequired(final int access,
       final String name, final String desc, final String signature,
       final String[] exceptions, final MethodVisitor methodVisitor) {
-
-    if (name.equals("<clinit>")) {
-      foundClinit = true;
-    }
 
     return new CoverageAnalyser(this, this.classId, this.probeCount,
         methodVisitor, access, name, desc, signature, exceptions);
@@ -100,57 +95,6 @@ public class CoverageClassVisitor extends MethodFilteringAdapter {
     super.visitField(Opcodes.ACC_STATIC | Opcodes.ACC_FINAL | Opcodes.ACC_PUBLIC
             | Opcodes.ACC_SYNTHETIC, CodeCoverageStore.PROBE_LENGTH_FIELD_NAME, "I",
         null, this.probeCount + 1);
-
-    //If there is no <clinit>, then generate one that sets the probe field directly
-    if (!foundClinit) {
-      MethodVisitor clinitMv = this.cv
-          .visitMethod(Opcodes.ACC_STATIC, "<clinit>", "()V", null, null);
-      clinitMv.visitCode();
-
-
-      pushConstant(clinitMv, this.classId);
-      pushConstant(clinitMv, this.probeCount);
-      clinitMv
-          .visitMethodInsn(Opcodes.INVOKESTATIC, CodeCoverageStore.CLASS_NAME,
-              "getOrRegisterClassProbes", "(II)[Z", false);
-
-      clinitMv.visitFieldInsn(Opcodes.PUTSTATIC, className,
-          CodeCoverageStore.PROBE_FIELD_NAME, "[Z");
-      clinitMv.visitInsn(Opcodes.RETURN);
-      clinitMv.visitMaxs(0, 0);
-      clinitMv.visitEnd();
-    }
-  }
-
-  private void pushConstant(MethodVisitor mv, int value) {
-    switch (value) {
-      case 0:
-        mv.visitInsn(Opcodes.ICONST_0);
-        break;
-      case 1:
-        mv.visitInsn(Opcodes.ICONST_1);
-        break;
-      case 2:
-        mv.visitInsn(Opcodes.ICONST_2);
-        break;
-      case 3:
-        mv.visitInsn(Opcodes.ICONST_3);
-        break;
-      case 4:
-        mv.visitInsn(Opcodes.ICONST_4);
-        break;
-      case 5:
-        mv.visitInsn(Opcodes.ICONST_5);
-        break;
-      default:
-        if (value <= Byte.MAX_VALUE) {
-          mv.visitIntInsn(Opcodes.BIPUSH, value);
-        } else if (value <= Short.MAX_VALUE) {
-          mv.visitIntInsn(Opcodes.SIPUSH, value);
-        } else {
-          mv.visitLdcInsn(value);
-        }
-    }
   }
 
   public String getClassName() {
--- a/pitest/src/main/java/org/pitest/coverage/CoverageClassVisitor.java
+++ b/pitest/src/main/java/org/pitest/coverage/CoverageClassVisitor.java
@@ -19,6 +19,7 @@ package org.pitest.coverage;
 
 import org.objectweb.asm.ClassWriter;
 import org.objectweb.asm.FieldVisitor;
+import org.objectweb.asm.Label;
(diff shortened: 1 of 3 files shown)


hcoles:
Hi @LaurentTho3 thanks for looking at this. 

Under most circumstances this sort of approach would not be acceptable as it is not thread safe, however in this case I **think** the worst consequence would be that we would loose some coverage as the array would be overwritten. That would be an improvement on erroring with a NullPointerException.

If you have time to investigate can you see if you can figure out why <clinit> is not called for #770 and #746? A solution that robustly ensured this happened would be preferable.

LaurentTho3:
Hello @hcoles ,

Sorry, I did not think about the thread-safety aspect of things.

I do not completely understand the situation, but from what I understand, the problem is not that clinit is not called, but that it is not called first. This post seems to be related: https://stackoverflow.com/q/11419519 .

The problem would then occur when a class has a static member of its own type I think. Looking at the (instrumented) bytecode generated for com/example/NestedClassTest$MyEnum$1 in #770 would confirm this:
<details>
  <summary>Bytecode</summary>
// class version 52.0 (52)
// access flags 0x4030
final enum com/example/NestedClassTest$MyEnum$1 extends com/example/NestedClassTest$MyEnum {

  OUTERCLASS com/example/NestedClassTest$MyEnum null
  // access flags 0x4009
  public static enum INNERCLASS com/example/NestedClassTest$MyEnum com/example/NestedClassTest MyEnum
  // access flags 0x4008
  static enum INNERCLASS com/example/NestedClassTest$MyEnum$1 null null
  // access flags 0x1008
  static synthetic INNERCLASS com/example/NestedClassTest$1 null null

  // access flags 0x1019
  public final static synthetic [Z $$pitCoverageProbes

  // access flags 0x1019
  public final static synthetic I $$pitCoverageProbeSize = 5

  // access flags 0x0
  <init>(Ljava/lang/String;I)V
    GETSTATIC com/example/NestedClassTest$MyEnum$1.$$pitCoverageProbes : [Z
    IFNONNULL L0
    ICONST_3
    GETSTATIC com/example/NestedClassTest$MyEnum$1.$$pitCoverageProbeSize : I
    INVOKESTATIC sun/pitest/CodeCoverageStore.getOrRegisterClassProbes (II)[Z
    PUTSTATIC com/example/NestedClassTest$MyEnum$1.$$pitCoverageProbes : [Z
   L0
   FRAME FULL [U java/lang/String I] []
    GETSTATIC com/example/NestedClassTest$MyEnum$1.$$pitCoverageProbes : [Z
    DUP
    ICONST_0
    ICONST_1
    BASTORE
    ASTORE 3
    ALOAD 3
    ICONST_1
    ICONST_1
    BASTORE
    ALOAD 0
    ALOAD 1
    ILOAD 2
    ACONST_NULL
    INVOKESPECIAL com/example/NestedClassTest$MyEnum.<init> (Ljava/lang/String;ILcom/example/NestedClassTest$1;)V
    ALOAD 3
    ICONST_2
    ICONST_1
    BASTORE
    RETURN
    MAXSTACK = 4
    MAXLOCALS = 4

  // access flags 0x8
  static <clinit>()V
    GETSTATIC com/example/NestedClassTest$MyEnum$1.$$pitCoverageProbes : [Z
    IFNONNULL L0
    ICONST_3
    ICONST_4
    INVOKESTATIC sun/pitest/CodeCoverageStore.getOrRegisterClassProbes (II)[Z
    PUTSTATIC com/example/NestedClassTest$MyEnum$1.$$pitCoverageProbes : [Z
   L0
   FRAME FULL [] []
    RETURN
    MAXSTACK = 2
    MAXLOCALS = 0
}
</details>

Here ```static enum INNERCLASS com/example/NestedClassTest$MyEnum$1 null null``` would seem to be the culprit.

Similarly, looking at the Type class which was one of the classes giving errors in #746 (https://github.com/apache/commons-bcel/blob/rel/commons-bcel-6.4.0/src/main/java/org/apache/bcel/generic/Type.java) we see the following static member: ```public static final Type UNKNOWN = new Type(Const.T_UNKNOWN, "<unknown object>") { };```

There might be other cases where this problem might occur, but I do not know which ones.

A solution might be to somehow ensure that clinit comes first as it seems that the initialisers are run "in textual order", but that might create other problems. Otherwise, I do not see another workaround for this problem, although I will keep thinking about it.

hcoles:
I'm now wondering if there is a hole in the static initalizer approach when classes are loaded by reflection with initalize = false passed in.

Perhaps your lazy initialization route is the right way to go, but we'd need to handle the thread safety concern. I don't think a synchronized block on every probe retrieval would be acceptable, but we could possibly follow the double checked locking pattern.

https://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html

This would require that the field was volatile though, and I have a memory of that being an issue. There are at least some scenarios where I think the jvm spec says that it can't be (interfaces?).

Might be that we need different strategies for different types of class.

LaurentTho3:
Loading classes with reflection and initialize = false would cause the same problem I guess, yes.

The java specs say that "It is a compile-time error if a final variable is also declared volatile." (https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.3.1.4), and also that "Every field declaration in the body of an interface is implicitly public, static, and final" (https://docs.oracle.com/javase/specs/jls/se8/html/jls-9.html#jls-9.2), so it looks like volatile and interfaces would not mix. Could we make the field not final as we are generating synthetic bytecode?

I have started playing a bit with the double checked locking pattern. I use the following code for the synchronised code but it fails in MutationCoverageReportSystemTest.shouldMutateClassesSuppliedToAlternateClassPath, I guess because Type.getObjectType(className) is not finding the classes not on the classpath. I will try to think of a better way.
 ```
if ((access & Opcodes.ACC_STATIC) != 0) { //method is static
      this.mv.visitLdcInsn(Type.getObjectType(className)); //class
    } else {
      this.mv.visitVarInsn(ALOAD,0); //this
    }
    this.mv.visitInsn(MONITORENTER);
```

**Comment by LaurentTho3** [comment]
Reading up on lazy initialisation, and I was wondering if the on demand holder idiom https://en.wikipedia.org/wiki/Initialization-on-demand_holder_idiom could be a solution? At least it does not require volatile, but maybe breaks other things I did not think about.

**Comment by LaurentTho3** [comment]
Tried my hand at the double locking but it fails in PitMojoIT.shouldWorkWithGWTMockito because of the following error: ```java.lang.ClassNotFoundException: boolean[] at java.base/java.lang.Class.forName0(Native Method)``` thrown in ```sample.MyWidgetTest```.

I really do not get why this error is popping up from the changes I made, and it seems to not be happening on all jdk versions. Maybe I am missing something obvious?

**Comment by hcoles** [comment]
Not clear on what is happening here.

The jdk version dependent behaviour I do understand - gtwmockito doesn't work with all versions, so the test only runs for those it is supported for.

Had a bit of a hack locally with your branch.

The method init probes method ought to be marked private and synthetic (so that reflection frameworks don't pick it up). Changing this didn't make gwtmockito any happier.

On a hunch I hacked it so that the probes were always initialized in the static initialiser for classes as well as interfaces. This did work, although i don't really understand why.

So that seems to be a working solution, but given the lazy initialisation will only be used in corner cases it doesn't feel great. 

An alternative would be to dig into why the initializer wasn't called in the two issues that resulted in null pointers, and add whatever code is required for those scenarios (plus check whether class.forname with initialize false passed in does cause and issue).

**Comment by hcoles** [comment]
@LaurentTho3 Finally understand what's going on here. The issue is that parent classes are always initilialized before their children. So given something like 

```java
public class ParentChildInitializationTest {
    @Test
    public void test() {
        new TesteeChild();
    }
}

class TesteeParent {
    static TesteeChild child = new TesteeChild();
}

class TesteeChild extends TesteeParent {
    final static Object f = "hello";

    TesteeChild() {
        System.out.println(f);
    }
}
```

You get output like

```
null
hello
```

so the array probe field in child classes like this are not initialized when their init method is called from the parent static initializer.

Knowing this, I think the appropriate fix is your initial null check, but with the static initializer left in place. In 99% of cases the probe will be created as normal, for the corner case it will be created by the init call from the parent static initializer.

I've put this together in #850.

Thanks for your work on this, I suspect it has been blowing up in a lot of codebases without being reported.

Issue #128, opened 2014-06-09
Break build on low test strength


hcoles:
By popular demand pitest makes it possible to break the build if a certain line coverage or mutation score is not achieved.

Using a metric in this way can however drive the wrong behaviors, resulting in over specified tests and unit tests for code that would be more sensibly integration tested rather than unit tested.

A better metric on which to break the build might be a measure of average test strength - the % of mutants that are executed by a test, but not killed by it.

This could be calculated from the number of killed mutants dived by the number of killed and surviving mutants (excluded those with a status of uncovered).

This ought to provide a metric that falls weak tests are introduced.

hcoles:
Closed by #774

PR #774, opened 2020-06-18
Add test strength statistic

alex859:
Hello @hcoles, first of all thanks a lot for this project!

This PR is to address enhancement #128 as I would find it useful as well.


hcoles:
Thanks @alex859, been meaning to put this in for a long time.

PR #638, opened 2019-08-26
Do not report 100% if not all mutants were killed

Vampire:

Issue #375, opened 2017-07-17
mutationThreshold as 100 is not guaranty that all classes is covered by 100%


romani:
steps to reproduce
```
$ git clone git@github.com:checkstyle/checkstyle.git
$ git checkout 3df996acc3bfc8072bcb5ce50009add08c3104dc
$ mvn -Ppitest-checks-annotation,no-validations clean verify org.pitest:pitest-maven:mutationCoverage -Dverbose=true
```

![image](https://user-images.githubusercontent.com/812984/28281515-3cd06fda-6adc-11e7-8ad2-e85e43435e7c.png)

Unfortunately it could lead to leek of incomplete test even on CI level is demanded to be 100% for Mutation Coverage `<mutationThreshold>100</mutationThreshold>` .

Expected: when some class is less then 100%, summary mutation coverage should be less then 100% and to have build failed if `<mutationThreshold>100</mutationThreshold>` present in config.

romani:
the same problem for `coverageThreshold`, 100 in config does not trigger build failure when come class is covered a bit lower than 100.
Post processing of report by "grep/...." is required to enforce 100 always, and have no leak.

nicolasSoyeur:
@romani I have the same issue sometimes. My team is alway putting the threshold to 100. I found a round in the source code of the plugin, which put the percentage to 100% even if you have mutant still alive.
e.g : you have 500 mutant generated, but 499 killed, the pluging compute a Math.round((100 / 500) * 499) = Math.round(99.8) = 100

 I don't know for the coverageeThreshold. I will check later

romani:
Is it same as https://github.com/hcoles/pitest/pull/638 ?

nicolasSoyeur:
Ohh sounds like there are conflict about rounding down and up :(

romani:
@Kevin222004, please send PR to update pitest to latest version

Vampire:
Ah, no, now I can say my PR was not related. :-D
But probably the same logic should be used:
![image](https://github.com/hcoles/pitest/assets/325196/160061c7-9fcf-4243-966a-99a13f9d611a)

Vampire:
#1221 should finally fix it hopefully though

Vampire:
It was merged, so next release after 1.14.1 should fix it.

PR #776, opened 2020-06-23
Rounding can give false 100% in html report

esovan:
To avoid false 100% coverage in html report i suggest using floor instead of round

### DIFF
--- a/pitest-html-report/src/main/java/org/pitest/mutationtest/report/html/MutationTotals.java
+++ b/pitest-html-report/src/main/java/org/pitest/mutationtest/report/html/MutationTotals.java
@@ -49,13 +49,13 @@ public class MutationTotals {
   }
 
   public int getLineCoverage() {
-    return this.numberOfLines == 0 ? 100 : Math
+    return this.numberOfLines == 0 ? 100 : (int) Math
         .floor((100f * this.numberOfLinesCovered) / this.numberOfLines);
   }
 
   public int getMutationCoverage() {
     return this.numberOfMutations == 0 ? 100
-        : Math.floor((100f * this.numberOfMutationsDetected)
+        : (int) Math.floor((100f * this.numberOfMutationsDetected)
             / this.numberOfMutations);
   }
 
--- a/pitest-html-report/src/main/java/org/pitest/mutationtest/report/html/MutationTotals.java
+++ b/pitest-html-report/src/main/java/org/pitest/mutationtest/report/html/MutationTotals.java
@@ -50,12 +50,12 @@ public class MutationTotals {
 
   public int getLineCoverage() {
     return this.numberOfLines == 0 ? 100 : Math
-        .round((100f * this.numberOfLinesCovered) / this.numberOfLines);
+        .floor((100f * this.numberOfLinesCovered) / this.numberOfLines);
   }
 
   public int getMutationCoverage() {
     return this.numberOfMutations == 0 ? 100
-        : Math.round((100f * this.numberOfMutationsDetected)
+        : Math.floor((100f * this.numberOfMutationsDetected)
             / this.numberOfMutations);
   }
 
--- a/pitest-html-report/src/test/java/org/pitest/mutationtest/report/html/MutationTotalsTest.java
+++ b/pitest-html-report/src/test/java/org/pitest/mutationtest/report/html/MutationTotalsTest.java
@@ -36,7 +36,7 @@ public class MutationTotalsTest {
   public void shouldCorrectlyCalculateLineCoverageWhenPartiallyCovered() {
     this.testee.addLines(63);
     this.testee.addLinesCovered(20);
-    assertEquals(32, this.testee.getLineCoverage());
+    assertEquals(31, this.testee.getLineCoverage());
   }
 
   @Test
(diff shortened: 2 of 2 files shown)


arnobl:
I also faced this issue in my project https://github.com/interacto/interacto-java-api/releases in its version 4.3.0.
One mutant was not killed but the score was 100%. My mutation threshold is 100% but pitest builds. 
Maybe the score should not be an int value.

romani:
@esovan , is this PR to fix problem https://github.com/hcoles/pitest/issues/375 ?

esovan:
i think it will fix it yes.
The problem i set out to fix was that we could get 100% in the total value, but not in all packages. Example in following picture:
image

Vampire:
Imho rounding is the better way generally, but it should ensure that it is never rounded to 100%.

That's why I opened https://github.com/hcoles/pitest/pull/638 two years ago that fixes exactly that and is conflicting to this change.

esovan:
In my opinion it should never be allowed to round up. Because then it is promising lets say 80% is covered, even though it might just be 79.5%.
With this change it will promise only 79% and you get the last .5% as a bonus.
I don't see why you would be interested in the coverage if its not to say: "this amount of the code is covered"

hcoles:
Sorry I've been quiet/absent in these discussions. I don't pay much attention to the % myself (I use pitest purely for qualative analysis), but every time the calculations are touched it causes disagreement. 

If you scan through the issues you'll find conflicting complaints that it should round up vs down.

If someone can put together a PR that makes it an configuration parameter (and defaults it whichever way they prefer) I'll merge it in.

pitest-html-report/src/main/java/org/pitest/mutationtest/report/html/MutationTotals.java
```diff
@@ -49,13 +49,13 @@ public void addMutationsDetetcted(final long mutationsKilled) {
   }
 
   public int getLineCoverage() {
-    return this.numberOfLines == 0 ? 100 : Math
-        .round((100f * this.numberOfLinesCovered) / this.numberOfLines);
+    return this.numberOfLines == 0 ? 100 : (int) Math
+        .floor((100f * this.numberOfLinesCovered) / this.numberOfLines);
```
nicolasSoyeur, line 53: you don't even need to do a floor. if you cast the result into an int, it will "floor" it also. But maybe the floor is more explicit on what it is done here.

(int) ((100f * this.numberOfLinesCovered) / this.numberOfLines);

nicolasSoyeur:
#776 is correcting the html report
#638 is correcting the percentage computation used by the build (it means, it will fail the build if you have a threshold = 100 % and you still have one mutant alive)

We can probably put the 2 PR in one.

nicolasSoyeur:
This issue "the build is not failing if I have one mutant still allive and my threshold = 100%" is very painfull to my team. We are facing this issue multiple times a month, because our threshold is to 100%.
I do agree with @esovan, the round floor for the percentage computation is the more logical way of computing it.

If I have a mutant still alive, Im not into "100 mutation coverage". It is the principile of the mutation test : 
you want to check if your unit test are bullet proof or not. If you have a mutation coverage of 99.6%, you are not 100% bullet proof :)

romani:
Off topic:

>for qualative analysis

@hcoles , Can you share details ?

hcoles:
@romani Posh way of saying I look at the detailed results (generated frequently) rather than the summary figures. See https://blog.pitest.org/dont-let-your-code-dry/

romani:
thanks a lot for sharing an article.

Right now I understand why you do not understand us. You work in different environment, where coverage is not enforced on fanatic level 100% and other engineers might not be interested to use this tool and care less on testing.

In Checkstyle project we are fanatic on quality and we have 100% branch coverage and almost reached to 100% by pitest on whole code. It helped us tremendously to prove that there is no extra code(code with no effect behavior) in contribution by PR. It helps to let contributor to think on tests more, so as result our code is more likely to stay

willhains:
I think the problem is when coverage/mutation percentage is used as a threshold for passing builds. The rounded percentage score under-reports missed coverage/mutations, so when used as a threshold, builds may pass despite failure to meet the minimum threshold.

How about adding "missed coverage" and "missed mutation" thresholds, expressed as a *maximum count*? This [is what JaCoCo does](https://www.eclemma.org/jacoco/trunk/doc/check-mojo.html), and it works very well for teams that want 100% perfect scores.

Something like:
```xml
<execution>
  <id>mutation-coverage</id>
  <goals>
    <goal>mutationCoverage</goal>
  </goals>
  <configuration>
    <missedCoverageThreshold>0</missedCoverageThreshold>
    <missedMutationThreshold>0</missedMutationThreshold>
  </configuration>
</execution>
```

nicolasSoyeur:
Im trying to add the configuration parameters that @hcoles proposed. I think I did it well for the maven build, but I need to work on the "reporting phase"
you can have a look at it here : https://github.com/nicolasSoyeur/pitest/tree/nso/roundUpMutationTotal

Vampire:
@esovan you could consider whether this PR is obsolete.
My #638 was finally merged into 1.13.2, so you should not get false 100% anymore.
It still uses `round`, but uses at max 99% as long as there is at least one remaining mutant.

Vampire:
Ah, no, those are different numbers. But I suggest the same logic, properly rounded but at most 99% if there is at least one left.

Vampire:
Ok, now this should be obsolete :-)

esovan:
seems, the consensus is that this is resolved. I haven't tested this myself but i trust the community :)

Issue #464, opened 2018-05-02
Cannot pass jvmArgs through command-line in maven plugin


sabaka:
Hi,

I want to pass some property to project with ``-Ddo.smt.ex=doThis``.
As I understood it's possible to do only using ``jvmArgs``, but I can't pass it from command line. Looks like it works only when set in pom.xml.

I think this is due to omitted ``property="jvmArgs"`` on ``@Parameter`` annotation in mojo here: https://github.com/hcoles/pitest/blob/master/pitest-maven/src/main/java/org/pitest/maven/AbstractPitMojo.java#L174

PR #798, opened 2020-09-19
Add missing property for jvmArgs parameter #464

yfrolov:
Hi,

I would like to suggest a fix for the issue #464 when `jvmArgs` property is not being picked up by `pitest-maven-plugin` if passed through the CLI.

For example, in the case when I can't make any changes in `pom.xml` I would like to be able to supply the argument (e.g. `-DjvmArgs=-ea`) through the CLI and expect it to be injected into the plugin. 

Thank you,
Yuriy


hcoles:
thanks!

yfrolov:
Hello @hcoles,

Do you think this pull request should be linked to the issue #464 ?

Regards,
Yuriy

PR #799, opened 2020-09-19
fix input/ouput file charset issue

qxo:
ref: https://openjdk.java.net/jeps/8187041

https://stackoverflow.com/questions/696626/java-filereader-encoding-issue
In earlier versions of java, you need to use new InputStreamReader(new FileInputStream(pathToFile), <encoding>).


hcoles:
Thanks for this, it wouid be good to be able to set the encodings.

However i'm not keen on the mutable static variable + setting it via a system property. Could you update to instead pass the data explicitly? This would require adding a filed to the ListenerArguments class.

qxo:
> Thanks for this, it wouid be good to be able to set the encodings.
> 
> However i'm not keen on the mutable static variable + setting it via a system property. Could you update to instead pass the data explicitly? This would require adding a filed to the ListenerArguments class.

How ? Please give me a example:)

PR #847, opened 2020-12-24
filter mutation results for multiple (generated) methods in same LOC

mhoennig:
avoidsa false positives stemming from generated code from Kotlin by-delegation


mhoennig:
created in wrong repo, sorry

Issue #596, opened 2019-04-12
Broken stack traces


Pr0methean:
While trying to debug my Pitest configuration, I got the following in my log:
```
7:53:28 PM PIT >> INFO : MINION : .internal.loader.BuiltinClassLoader.defineClass(BuiltinClassLoader.java:802)
	at java.base/jdk.internal.loader.BuiltinClassLoader.findClassOnClassPathOrNull(BuiltinClassLoader.java:700)
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClassOrNull(B
7:53:28 PM PIT >> INFO : MINION : uiltinClassLoader.java:623)
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:581)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
	at java.base/java.lang.ClassLoader.loadC
```
An option to output these chopped-up stack traces (whether to the log or into a file in `target/`) in one piece would be very helpful.

PR #857, opened 2021-01-16
Pass data from stdout and stderr of spawned processes line by line to…

kgeilmann:
… the consumer - Fix for #596.

This ensures, that the prefixes added by a consumer do not occur in the middle of the message, which happened in the old implementation if the message was longer than 256 bytes (often the case for stacktraces).


hcoles:
Thanks!

Issue #269, opened 2016-06-02
PIT 1.1.10 maxSurviving issues


tbuchber:
I tested PIT 1.1.10. I love that I can now set maxSurviving. However, I did find a couple of issues.

1.The minimum maxSurviving value supported is 1. I really want to set it to 0. However, line 387 in pitest/AbstractPitMojo.java [if ((this.maxSurviving > 0)] limits maxSurviving to >=1.

2.The message at failure is displaying the incorrect number of max survivors. Line 391 in pitest/AbstractPitMojo.java is displaying mutationThreshold instead of maxSurviving.

theangrydev:
Would setting <mutationThreshold>100</mutationThreshold> achieve what you want instead?

hcoles:
@theangrydev maxSurviving was introduced as the threshold doesn't quite allow you the break the build on a surviving mutant due to rounding.

It sounds like it is impressively broken for such as simple feature. I'll put out a fix when I get chance, or am happy to accept a PR.

tbuchber:
Your fixes look correct to me. Thanks for the fixes. I look forward to trying 1.1.11. Thank you!

PR #858, opened 2021-01-22
Fixed tests not failing when maxSurviving=0 and typo

alexkoltz:
Currently when maxSurviving is set to 0, an error will still not be thrown due to this restriction `if ((threshold > 0)` in src/main/java/org/pitest/mutationtest/commandline/MutationCoverageReport.java in `throwErrorIfMoreThanMaxSuvivingMutants`.

We can safely set it to `if ((threshold >= 0)` as the default is -1.

This should finish the fixes that were started to address this issue: [https://github.com/hcoles/pitest/issues/269](https://github.com/hcoles/pitest/issues/269).

Also fixed a typo in that method's name changing `throwErrorIfMoreThanMaxSuvivingMutants` to `throwErrorIfMoreThanMaxSurvivingMutants`.

Let me know if any additional information/clarification from me would be helpful.

Thanks!


hcoles:
Thanks!

PR #877, opened 2021-03-26
Add pitest-aggregate api for aggragating XML report

iadcode:
### Change Background
I have a multi-module project and was wanted to create an aggregated XML report of all my module for sonar. I have chosen to not set this project up using submodule in sonar due to various reasons and as such I need a single mutation report. I notice that the report aggregator only does aggregated HTML report at the moment. While aggregating the XML reports could simply be done b reading these reports in as xml objects, combining them and rewriting them, I feel like doing this using the Listeners would produce a more consistent format with what pitest is producing.

### Change Content
Simple change based off the original function to aggregate HTML, but using the XMLReportListener instead of the MutationHtmlReportListener.

### DIFF
--- a/pitest-maven/src/main/java/org/pitest/maven/report/AbstractPitAggregationReportMojo.java
+++ b/pitest-maven/src/main/java/org/pitest/maven/report/AbstractPitAggregationReportMojo.java
@@ -63,6 +63,7 @@ abstract class AbstractPitAggregationReportMojo extends PitReportMojo {
           .build();
 
       reportAggregator.aggregateReport();
+      reportAggregator.aggregateXmlReport(false);
     } catch (final Exception e) {
       throw new MavenReportException(e.getMessage(), e);
     }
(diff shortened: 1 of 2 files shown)


hcoles:
This adds a method, but the functionality doesn't seem to be exposed anywhere. Am I missing something?

iadcode:
@hcoles , I intended for it to be used the same way the aggregateReport function would be used as seen in the example given here https://pitest.org/aggregating_tests_across_modules.

At the moment, I've pulled the pitest-aggregator package into my version of the sbt-pit plugin to do the aggregation for HTML and hoping this request would go through so I will be able to use the function introduced to do the XML aggregation.

iadcode:
After replying to you, I reviewed the code and realised that the pitest maven plugin already has the aggregation built-in. I've updated the plugin to also aggregate the XML. If you think this is unreasonable, I can revert this. Would just be a nice to have for me, otherwise, I'll go back to writing my own aggregation plugin for my maven projects. Many thanks for taking the time to look at this.

pitest-maven/src/main/java/org/pitest/maven/report/AbstractPitAggregationReportMojo.java
```diff
@@ -63,6 +63,7 @@ protected void executeReport(final Locale locale)
           .build();
 
       reportAggregator.aggregateReport();
+      reportAggregator.aggregateXmlReport(false);
```
szpak, line 66: Thinking about the default login in the Gradle plugin. Should the XML report generated by default? Only if XML report is configured for the PIT "single" report? Or there should be a separate property for the aggregation mojo? WDYT?

PR #891, opened 2021-04-29
History performance

hcoles:
Writing history information is stupidly slow for larger codebases as it collates class level coverage data from instruction level data with OMG complexity.

This change drives the history off of the legacy line coverage information, which was previously lazily calculated. Since a recent change this information is also used by the report aggregation system. To prevent accidental breakage by using data not present at aggregate time, the line coverage data has been hived off into a seperate class.

This horrific mess and needs revisting.


hcoles:
An example project analysis time before = 1 hr, after = 45 min.

PR #897, opened 2021-05-09
fix mutation message (emptyList -> emptySet)

### LINKED ISSUE #895: Wrong message: "replaced return value with Collections.emptyList" should be "emptySet"
https://github.com/hcoles/pitest/blob/5962ff1473c10e8e2cf98559b22969f7f0d664e8/pitest/src/main/java/org/pitest/mutationtest/engine/gregor/mutators/EmptyObjectReturnValsMutator.java#L217

Should be emptySet, not emptyList

Related to #894

C-Otto:
fixes #895


hcoles:
thanks!

Issue #910, opened 2021-06-07
Pitest should add all dependencies to forked classpath


A248:
Using JUnit 5 and Maven surefire, it is common to declare the dependency on JUnit-Jupiter-Engine separately, so that test classes can't depend on Jupiter-Engine classes.

This is explained at https://maven.apache.org/surefire/maven-surefire-plugin/examples/junit-platform.html . The example described there is:

```xml
<dependencies>
    [...]
    <dependency>
        <groupId>org.junit.jupiter</groupId>
        <artifactId>junit-jupiter-api</artifactId>
        <version>5.3.0</version>
        <scope>test</scope>
    </dependency>
    [...]
</dependencies>
...
<build>
    <plugins>
        [...]
        <plugin>
            <groupId>org.apache.maven.plugins</groupId>
            <artifactId>maven-surefire-plugin</artifactId>
            <version>3.0.0-M5</version>
            <dependencies>
                <dependency>
                    <groupId>org.junit.jupiter</groupId>
                    <artifactId>junit-jupiter-engine</artifactId>
                    <version>5.3.2</version>
                </dependency>
            </dependencies>
        </plugin>
        [...]
    </plugins>
</build>
```

### Relation to Pitest

This doesn't work with Pitest-maven. Pitest-maven collects the dependencies for itself and for the test-classpath.

However, Pitest-maven filters the forked classpath based on which dependencies are pitest plugins. This removes declarations of JUnit-Jupiter-Engine.

For example, use the following configuration:

```xml
			<plugin>
				<groupId>org.pitest</groupId>
				<artifactId>pitest-maven</artifactId>
				<version>1.6.7</version>
				<dependencies>
					<dependency>
						<groupId>org.pitest</groupId>
						<artifactId>pitest-junit5-plugin</artifactId>
						<version>0.14</version>
					</dependency>
					<dependency>
						<groupId>org.junit.jupiter</groupId>
						<artifactId>junit-jupiter-engine</artifactId>
						<version>${junit.version}</version>
					</dependency>
				</dependencies>
			</plugin>
```

which results in:

```
[INFO] Found shared classpath plugin : Default mutation engine
[INFO] Found shared classpath plugin : JUnit 5 test framework support
[INFO] Found shared classpath plugin : JUnit plugin
[INFO] Found shared classpath plugin : TestNG plugin
[INFO] Adding org.pitest:pitest-junit5-plugin to SUT classpath
[INFO] Adding org.pitest:pitest to SUT classpath
[...]

PR #911, opened 2021-06-07
Use all plugin dependencies

A248:
Solves #910 

Allows adding dependencies like JUnit-Jupiter-Engine to pitest-maven's dependencies without having to add them to the project test classpath.


A248:
This adds a lot of maven-related dependencies to the SUT classpath, so looks like I'll need to fix that.

A248:
I can't find a reasonable way to do this but someone else is welcome to try

PR #919, opened 2021-08-05
Filter junk mutations from java records

hcoles:


pitest-entry/src/main/java/org/pitest/mutationtest/build/intercept/javafeatures/RecordFilter.java
```diff
@@ -0,0 +1,61 @@
...
+    private ClassTree currentClass;
+
+    @Override
+    public InterceptorType type() {
+        return InterceptorType.FILTER;
+    }
+
+    @Override
+    public void begin(ClassTree clazz) {
+        isRecord = "java/lang/Record".equals(clazz.rawNode().superName);
```
TomRK1089, line 27: Just FYI, the `Class<T>` type also provides a simple flag to indicate if a type is a record without checking the superclass name: https://docs.oracle.com/en/java/javase/16/docs/api/java.base/java/lang/Class.html#isRecord() not sure if that's easier to use here or if it can be exposed by the `ClassTree` type.

hcoles: We need to compile and run under java 8, is isRecord is not available.

PR #923, opened 2021-08-16
Refactorings that may break plugin compatibility

hcoles:


szpak:
It ran it with the functional tests in gradle-pitest-plugin and it all passed (except non-related https://github.com/hcoles/pitest-plugins/issues/2 and https://github.com/hcoles/pitest-plugins/pull/4). Just writing this comment I realized that you were writing about the other PIT plugins :).

hcoles:
Thanks @szpak you've done a lot more checking than I've done.

The idea here was to clean up and simplify a few things that have been bugging me for a while that requires changes to classes used in various interfaces. The boundary of where the plugin api starts and ends isn't very well defined, but it is possible that the changes will break something somewhere so will come with a version bump to 1.7.0. Fortunately there aren't that many plugins out there.

Since the version is going to be bumped I may make some other breaking changes in the 1.7.0 release that might affect the gradle plugin. Only thing that comes to mind at the moment is stripping out some of the defunct parameters, but it is possible there might be other changes.

PR #934, opened 2021-09-17
Configure console verbosity

hcoles:


pitest/src/main/java/org/pitest/coverage/execute/CoverageOptions.java
```diff
@@ -31,17 +32,17 @@
```
davidburstrom, line 32: I suppose technically, serialVersionUID should be updated :)

pitest-maven/src/main/java/org/pitest/maven/AbstractPitMojo.java
```diff
@@ -630,6 +635,7 @@ public boolean isExportLineCoverage() {
     return this.exportLineCoverage;
   }
 
+
```
davidburstrom, line 638: Accidental newline?

pitest-maven/src/main/java/org/pitest/maven/MojoToReportOptionsConverter.java
```diff
@@ -154,14 +155,23 @@ private ReportOptions parseReportOptions(final List<String> classPath) {
     return data;
   }
 
+  private void configureVerbosity(ReportOptions data) {
+    if (this.mojo.isVerbose()) {
+      data.setVerbosity(Verbosity.VERBOSE);
+    } else {
+      Verbosity v = Verbosity.fromString(mojo.getVerbosity());
```
davidburstrom, line 162: Am I thinking right that if the configuration explicitly says "verbose = false" but "verbosity=verbose", it will be verbose, and if it says "verbose = true" but "verbosity=QUIET" it will be also be verbose? Should it be considered an error to explicitly set both?

hcoles: Will probably remove the verbose option in a couple of releases time. In the meantime this approach lets people upgrade without changing their build scripts. 

Don't think the argument that this is better than erroring is particularly strong, but I don't think the counter argument is any stronger. Ultimately both behaviours are sensible, but arbtrary so makes sense to stick with the one that's already coded.

PR #965, opened 2021-12-13
BigDecimal mutator

MarcinNowak-codes:
Hi @hcoles,

Thanks for outstanding project Pitest. I’m using it for years. Pitest let think outside the box about problem I'm workin on. 

I'm often doing calculation based on BigDecimal. My implementation is similar to BigIntegerMutator. I read conversation in https://github.com/hcoles/pitest/pull/518#issue-361311895, but I couldn't find reason to not implement BigDecimal mutator.

Could you review? Suggestions are always welcome.

Kindly regards,

### DIFF
--- a/pitest/src/main/java/org/pitest/mutationtest/engine/gregor/mutators/experimental/BigDecimalMutator.java
+++ b/pitest/src/main/java/org/pitest/mutationtest/engine/gregor/mutators/experimental/BigDecimalMutator.java
@@ -1,0 +1,167 @@
+package org.pitest.mutationtest.engine.gregor.mutators.experimental;
+
+import org.objectweb.asm.Handle;
+import org.objectweb.asm.MethodVisitor;
+import org.objectweb.asm.Opcodes;
+import org.pitest.mutationtest.engine.MutationIdentifier;
+import org.pitest.mutationtest.engine.gregor.MethodInfo;
+import org.pitest.mutationtest.engine.gregor.MethodMutatorFactory;
+import org.pitest.mutationtest.engine.gregor.MutationContext;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+public enum BigDecimalMutator implements MethodMutatorFactory {
+  INSTANCE;
+
+  @Override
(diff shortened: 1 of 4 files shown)


pitest/src/main/java/org/pitest/mutationtest/engine/gregor/mutators/experimental/BigDecimalMutator.java
```diff
@@ -0,0 +1,167 @@
...
+      this.factory = factory;
+      this.context = context;
+    }
+
+    @Override
+    public void visitMethodInsn(int opcode, String owner, String name, String descriptor,
+        boolean isInterface) {
+      if (!owner.equals(expectedOwner) || opcode != Opcodes.INVOKEVIRTUAL) {
+        this.mv.visitMethodInsn(opcode, owner, name, descriptor, isInterface);
+        return;
```
hcoles, line 84: Returning without calling visitMethodInsn will corrupt the chain for mutators that come after this. This looks to also be a bug in the BigIntegeger mutator from which this has been copied.

The issue would be revealed if this mutator were enabled in a test with others that are called after it.

pitest/src/main/java/org/pitest/mutationtest/engine/gregor/mutators/experimental/BigDecimalMutator.java
```diff
@@ -0,0 +1,167 @@
...
+      if (replacement != null && replacement.descriptor.equals(descriptor)) {
+        MutationIdentifier identifier = context.registerMutation(factory, replacement.toString());
+        if (context.shouldMutate(identifier)) {
+          this.mv.visitMethodInsn(
+              opcode,
+              owner,
+              replacement.destinationName,
+              replacement.descriptor,
+              false);
+          return;
```
hcoles, line 97: Same issue as dicussed above.

hcoles:
@MarcinNowak-codes Ignore my earlier comments, they are incorrect. Merging.

MarcinNowak-codes:
Thank you @hcoles

PR #972, opened 2022-01-30
#750 -DskipTests Flag should be honoured in a maven run without additional configuration

### LINKED ISSUE #750: -DskipTests Flag not working
I know there is already a closed issue "Pit Maven Plugin Should Honor -DskipTests Flag #157", but the flag is not working, the command "mvn install -DskipTests" is still executing Pitest.

Affected Version org.pitest.pitest-maven.1.5.1

Btw.: The command "mvn install -DskipPitest"  works as expected!

marwin1991:
Currently when you have used:

```
mvn clean install -DskipTests
```
PiTest was executing tests, despite it should not, because people using this flag expect fast build without any tests.

Closes #750


pitest-maven-verification/src/test/java/org/pitest/PitMojoIT.java
```diff
@@ -353,17 +354,14 @@ public void shouldCorrectlyHandleOverrides() throws Exception {
         buildFilePath(testDir, "target", "site", "pit-reports").exists());
   }
 
-  @Test
+  @Test(expected = FileNotFoundException.class)
```
hcoles, line 357: This test checked if exclusions were read from surefire config, skipping tests only as a way of speeding the test up.

I think it needs to be updated to now ommit the`-DskipTests`, keeping the original assertion. A new test case is needed to ensure that the tests are now skipped when skipTests are set.

marwin1991: Okey, now I understand.

I will fix it with new commit :)

marwin1991: @hcoles can I ask you for another review :) ?

PR #625, opened 2019-06-19
fix(report-aggregate) Fix report aggregation

aurelien-baudet:
- #621: do not use test dependencies as test source directories
- #622: use the name of the mutator (it is not a class name written in mutations.xml)

PR #983, opened 2022-02-02
Mergeup of #625

therealryan:
This is the same changeset as the apparently-stalled #625, but with merge conflicts resolved

The conflicts turned out to be simple:

 * Catch blocks were collapsed in master. These had been deleted in the branch anyway
 * The `MethodName` class has been removed, so no need to call `MethodName.fromString()` (or `name()` in the tests)


hcoles:
Thanks @therealryan.

The tests around this area of the code are minimal, but that is an existing issue from when this feature was first added. If you can confirm that everthing works as expected when run against a real project I'll merge it in.

therealryan:
After:
1. checking out this branch
2. `mvn install`
3. updating my project to use the 1.0.0-SNAPSHOT version that was generated
4. Adding a `<value>XML</value>` element to `plugin/configuration/outputFormats`

I can confirm that report aggregation works as expected.

hcoles:
@therealryan Thanks, merged now.

therealryan:
Great stuff! All credit to @aurelien-baudet !

PR #993, opened 2022-03-03
remove pit rv operators

hcoles:
Removes the pit rv operators from the main codebase so they can instead be hosted in an optional plugin.

Although PIT RV is useful for the research community, including the operators in the main repo as proved a mainainance headache. New users activate them via the ALL group, then send e-mails questioning their behaviour.

Moving them to a seperate plugin will make it slightly harder for the research community to use them, but will prevent users who will not benefit from them activating them without understanding their pros and cons.


hcoles:
Migrated operators are available at https://github.com/pitest/pitest-rv-plugin

PR #996, opened 2022-03-08
reintroduce testPlugin parameter on command line

hcoles:
The gradle plugin seems to auto supply this parameter when not specified
by the user, resulting in an error.

Reintroducing for now, until gradle plugin is updated.


szpak:
To be tracked in https://github.com/szpak/gradle-pitest-plugin/issues/277.

Issue #675, opened 2019-09-18
jvmArgs comma separation breaks for certain JVM arguments


davidburstromspotify:
Given PIT 1.4.2 (and quite likely other versions as well) it is impossible to pass any JVM arguments that themselves have comma separation. Such an example is `-XX:CompileCommand=command,method[,option]` In our case it's necessary to affect how JIT is performed and thus it doesn't work.

plastic-karma:
We are having a similar problem in 1.6. We are trying to pass `-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=*:5005` to attach a debugger to the forked JVM, but it won't pick up the parts behind the comma correctly. It could be because of [OptionsParser](https://github.com/hcoles/pitest/blob/master/pitest-command-line/src/main/java/org/pitest/mutationtest/commandline/OptionsParser.java#L208) which assumes that the input is comma separated.

I wonder how this can be fixed. Changing the semantic for `jvmArgs` and not treating comma as a separator is backwards-incompatabile.

Could a new commandline `jvmArg` be introduced with just a single value that will not be parsed?

plastic-karma:
> Guess the only question is how it should behave if both parameters are supplied.

One option could be to just merge them all to a single list.

> If you've got time to put a pull request together and think about it, that would be great.

I'll do that. It is going to take a few days, but I am interested. :-)

PR #1007, opened 2022-04-01
allowing markers in jvm argument to preserve commas 

plastic-karma:
Introducing {} as markers to preserve commas in jvm arguments.

This fixes https://github.com/hcoles/pitest/issues/675.

### Testing
- added test cases
- ran pitest cli with `--jvmArgs -{agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=*:5005}` which lead to the child JVM waiting until a debugger was connected. Before this change, this failed.

### DIFF
--- a/pitest-command-line/src/test/java/org/pitest/mutationtest/commandline/CommaAwareArgsProcessorTest.java
+++ b/pitest-command-line/src/test/java/org/pitest/mutationtest/commandline/CommaAwareArgsProcessorTest.java
@@ -1,0 +1,59 @@
+package org.pitest.mutationtest.commandline;
+
+import joptsimple.ArgumentAcceptingOptionSpec;
+import joptsimple.OptionParser;
+import joptsimple.OptionSet;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.pitest.mutationtest.config.ConfigOption;
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+
+@RunWith(Parameterized.class)
+public class CommaAwareArgsProcessorTest {
+
+    @Parameterized.Parameters
+    public static Collection<Object[]> data() {
+        return Arrays.asList(new Object[][] {
+                {  Arrays.asList("--unrelated", "a"), Collections.emptyList() },
+                {  Arrays.asList("--jvmArgs", "a"), Arrays.asList("a") },
+                {  Arrays.asList("--jvmArgs", "a,b,c,d"), Arrays.asList("a","b","c","d") },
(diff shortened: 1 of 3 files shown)


hcoles:
Thanks @plastic-karma, I take it there was an issue with introducing a 2nd parameter?

plastic-karma:
> Thanks @plastic-karma, I take it there was an issue with introducing a 2nd parameter?

Actually no. Here is the trade off that I made:

Introduce new parameter: totally backwards compatible, but need to make changes in up streams like Grade, Ant and Maven

Alter existing parameter: minimally backwards incompatible (if someone uses brackets and commas in their arguments today it might break) but no upstream changes.

I opted to keep changes local. But I am curious what you think @hcoles - we can also introduce a new parameters and I can make changes in the upstreams.

hcoles:
The escaping is a bit of a hack, but I think it's a reasonable tradeoff.

Thanks for the PR, merging now.

pitest-command-line/src/test/java/org/pitest/mutationtest/commandline/CommaAwareArgsProcessorTest.java
```diff
@@ -0,0 +1,59 @@
...
+    public static Collection<Object[]> data() {
+        return Arrays.asList(new Object[][] {
+                {  Arrays.asList("--unrelated", "a"), Collections.emptyList() },
+                {  Arrays.asList("--jvmArgs", "a"), Arrays.asList("a") },
+                {  Arrays.asList("--jvmArgs", "a,b,c,d"), Arrays.asList("a","b","c","d") },
+                {  Arrays.asList("--jvmArgs", "a;b,c;d"), Arrays.asList("a;b","c;d") },
+                {  Arrays.asList("--jvmArgs", "{a,b,c,d}"), Arrays.asList("a,b,c,d") },
+                {  Arrays.asList("--jvmArgs", "{a;b;c;d}"), Arrays.asList("{a;b;c;d}") }, // no separators so region markers will not be removed
+                {  Arrays.asList("--jvmArgs", "{a,b},{c,d},{e,f}"), Arrays.asList("a,b","c,d","e,f") },
+                {  Arrays.asList("--jvmArgs", "{a,b@c,d}"), Arrays.asList("a,b@c,d") }, // pre-existing '@' will not be replaced with ","
```
szpak, line 30: Is it possible to quote the `{` usage inside the command?

hcoles: That hasn't been implemented. I agree it would be nice to have, but I think we can probably wait until someone encounters a scenario where it causes a problem.

As far as I'm aware '{' are valid (although I've not turned up anything to confirm/deny this with a quick google). Don't think I've ever encountered it in practice though.

PR #1014, opened 2022-04-29
check for junit4 on classpath before loading plugin

hcoles:


pitest-maven-verification/src/test/java/org/pitest/PitMojoIT.java
```diff
@@ -580,4 +580,15 @@ public void shouldNotNullPointerWhenEnumInitializerNotCalled() throws IOExceptio
...
   }
 
+  @Test
+  public void shouldFailCleanlyWhenTestPluginMissing() throws IOException, VerificationException {
+    File testDir = prepare("/pit-missing-test-plugin");
+    verifier.executeGoal("test");
+    verifier.executeGoal("org.pitest:pitest-maven:mutationCoverage");
+
+    String actual = readResults(testDir);
+    assertThat(actual).isNotEmpty();
```
szpak, line 590: Is there a way to test the error message or something else here?

hcoles: There isn't one at the moment - everything runs without finding tests as there are not applicable test plugins. Just pulling testng support out into an external plugin before fixing that.

PR #1015, opened 2022-04-29
Feature/remove testng

hcoles:


gmshake:
A little surprised that TestNG support is removed. Are there any clues about this?

hcoles:
@gmshake It's been exctracted into a plugin, available here -> https://github.com/pitest/pitest-testng-plugin

PR #1029, opened 2022-06-14
Support different input and output encodings

### LINKED ISSUE #637: Sourcefiles encoding option needed
Pitest needs a source file encoding option.
Currently always the system default encoding is used, which often is not correct.
It is quite common to have UTF-8 encoded source files and build on Windows.
this causes corrupted source display in the report.
For example in my copyright I have `Björn`, but it is displayed as `BjÃ¶rn`, which means it is saved in UTF-8, but read in ISO-8859-1.

The (or at least one) culprit is `org.pitest.mutationtest.tooling.DirectorySourceLocator.FileToReader#apply` as it opens a new `FileReader` without specifying an encoding.

hcoles:
Based on work by @qxo in #799

Finally fixes #637

### DIFF
--- a/pitest-aggregator/src/main/java/org/pitest/aggregate/ReportAggregator.java
+++ b/pitest-aggregator/src/main/java/org/pitest/aggregate/ReportAggregator.java
@@ -19,6 +19,7 @@ import org.pitest.util.Log;
 import org.pitest.util.ResultOutputStrategy;
 
 import java.io.File;
+import java.nio.charset.Charset;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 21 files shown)


pitest-aggregator/src/main/java/org/pitest/aggregate/ReportAggregator.java
```diff
@@ -132,6 +137,18 @@ public static class Builder {
     private final Set<File>      mutationResultsFiles    = new HashSet<>();
     private final Set<File>      sourceCodeDirectories   = new HashSet<>();
     private final Set<File>      compiledCodeDirectories = new HashSet<>();
+    private Charset inputCharset;
+    private Charset outputCharset;
+
+    public Builder inputCharSet(Charset inputCharset) {
```
szpak, line 143: @hcoles Working on the Gradle integration, I've noticed that you mix "charset" and "charSet" for fields, "setters" and getters. Maybe they all could switched to just "charset" (with deprecating the old methods)?

Issue #1032, opened 2022-06-15
 GC overhead limit exceeded


mgonzcast:
Hi,

  I have several modules in a project and only (the biggest one) fails with GC overhead limit exceeded. 

  I have searched around and I can´t find how to fix this. 

  Originally I run this command (which works in the rest of modules):

  `mvn clean install org.pitest:pitest-maven:mutationCoverage -DtimeoutConstant=10000 -DoutputFormats=XML -DexportLineCoverage=true -DtimestampedReports=false `

  and I get:

  `[INFO] Defaulting target classes to match packages in build directory
[INFO] Defaulting target tests to match packages in test build directory
11:38:57 PIT >> INFO : Verbose logging is disabled. If you encounter a problem, please enable it before reporting an issue.
11:39:04 PIT >> INFO : Incremental analysis reduced number of mutations by 0
11:39:04 PIT >> INFO : Created  668 mutation test units in pre scan
11:39:05 PIT >> INFO : Sending 861 test classes to minion
11:39:05 PIT >> INFO : Sent tests to minion
\11:43:12 PIT >> INFO : MINION : log4j:ERROR A "org.apache.log4j.xml.DOMConfigurator" object is not assignable to a "org.apache.log4j.spi.Configurator" variable.
11:43:12 PIT >> INFO : MINION : log4j:ERROR The class "org.apache.log4j.spi.Configurator" was loaded by
11:43:12 PIT >> INFO : MINION : log4j:ERROR [org.powermock.core.classloader.javassist.JavassistMockClassLoader@1e313166] whereas object of type
11:43:12 PIT >> INFO : MINION : log4j:ERROR "org.apache.log4j.xml.DOMConfigurator" was loaded by [sun.misc.Launcher$AppClassLoader@18b4aac2].
11:43:12 PIT >> INFO : MINION : log4j:ERROR Could not instantiate configurator [org.apache.log4j.xml.DOMConfigurator].
/11:59:21 PIT >> INFO : MINION : Java HotSpot(TM) 64-Bit Server VM warning: CodeCache is full. Compiler has been disabled.
11:59:21 PIT >> INFO : MINION : Java HotSpot(TM) 64-Bit Server VM warning: Try increasing the code cache size using -XX:ReservedCodeCacheSize=
-`

I have tried:

`
[...]

Issue #841, opened 2020-12-08
java.lang.OutOfMemoryError: Java heap space running against Kotlin project


JGulbronson:
Hello! I tried to run pitest against a Kotlin project, and it worked great for a small module. I tried on a larger one, and encountered the following error.

```
VM : OpenJDK 64-Bit Server VM
Vendor : Oracle Corporation
Version : 14.0.1+7
Uptime : 334777
Input ->
 1 : -Dfile.encoding=UTF-8
 2 : -Duser.country=CA
 3 : -Duser.language=en
 4 : -Duser.variant
BootClassPathSupported : false

java.util.concurrent.ExecutionException: java.lang.OutOfMemoryError: Java heap space
        at java.base/java.util.concurrent.FutureTask.report(FutureTask.java:122)
        at java.base/java.util.concurrent.FutureTask.get(FutureTask.java:191)
        at org.pitest.util.CommunicationThread.waitToFinish(CommunicationThread.java:59)
        at org.pitest.coverage.execute.CoverageProcess.waitToDie(CoverageProcess.java:35)
        at org.pitest.coverage.execute.DefaultCoverageGenerator.gatherCoverageData(DefaultCoverageGenerator.java:138)
        at org.pitest.coverage.execute.DefaultCoverageGenerator.calculateCoverage(DefaultCoverageGenerator.java:89)
        at org.pitest.coverage.execute.DefaultCoverageGenerator.calculateCoverage(DefaultCoverageGenerator.java:51)
        at org.pitest.mutationtest.tooling.MutationCoverage.runReport(MutationCoverage.java:115)
        at org.pitest.mutationtest.tooling.EntryPoint.execute(EntryPoint.java:121)
        at org.pitest.mutationtest.tooling.EntryPoint.execute(EntryPoint.java:51)
        at org.pitest.mutationtest.commandline.MutationCoverageReport.runReport(MutationCoverageReport.java:87)
        at org.pitest.mutationtest.commandline.MutationCoverageReport.main(MutationCoverageReport.java:45)
Caused by: java.lang.OutOfMemoryError: Java heap space
        at java.base/java.util.TreeMap.put(TreeMap.java:575)
        at java.base/java.util.TreeSet.add(TreeSet.java:255)
        at org.pitest.coverage.CoverageData.addTestsToBlockMap(CoverageData.java:145)
        at org.pitest.coverage.CoverageData.calculateClassCoverage(CoverageData.java:134)
[...]

PR #1035, opened 2022-06-16
Feature/coverage memory consumption

hcoles:
Reduces insanely high memory consumption by coverage data structures.

At least partially addressed #1032 and #841

The in memory coverage data store previously held a map entry for each instruction
within the code under test. This was almost unbelievably stupid and
resulted in very high memory consumption.

Coverage is tracked by block, which maps 1:many with instructions. The
reason coverage was being held against instruction seems to be because
the block wasn't being tracked when mutants were combined due code
inlining, so only the instruction could be used for targetting.

This change updates MutationDetails to track mutants affecting multiple
blocks, and changes the interface and internals of the coverage store
to be block driven.

This results in a change of interface to the core `MutationDetails` class, and a change 
in format to the xml output. This commit must therefore be accompanied by a version bump.

For the assertj project this reduces memory consumption from 3G to 1G.
Although a large improvement, this is still too high and will need to be looked at further.


vincentkupert:
Thank you for this improvement!

PR #1082, opened 2022-08-31
Fix inconsistent spacing in logging

davidburstrom:
This change is betting on that nobody is parsing the log output verbatim for further consumption.


hcoles:
thanks!

PR #1111, opened 2022-11-13
Fix log output

wangbo15:
The origin log output is like "100Executing 100 tests not run during discovery", and now the useless number at the very beginning is removed.


hcoles:
Thanks, that's slightly embarassing.

PR #1121, opened 2022-11-28
wip - remove gregor ClassInfo

hcoles:


pitest/src/main/java/org/pitest/mutationtest/engine/gregor/ClassInfo.java
```diff
@@ -27,10 +25,6 @@ public boolean isEnum() {
     return this.superName.equals("java/lang/Enum");
   }
 
-  public boolean isGroovyClass() {
-    return FArray.contains(this.interfaces, isAGroovyClass());
-  }
-
   private static Predicate<String> isAGroovyClass() {
     return a -> a.startsWith("groovy/lang/")
         || a.startsWith("org/codehaus/groovy/runtime");
```
szpak, line 30: I believe that private method could be removed as well.

szpak: Btw, it seems to work, but I wonder why you check the "interfaces" here? 

I would expect checking if the "top super class" is GroovyObject, but there is just a superName, so it might be harder at that level :-/.

hcoles: This change is still in progress, will hopefully end up with all of ClassInfo being removed. It's a long time since I put this groovy filtering in place, looking at the tests that interface stuff might be required to detect closures (at least in the version of groovy that existed at the point I first wrote this).

szpak: It might be even after you reported some issues with mutating Groovy code - yeaaaars ago :-). I missed the WIP status.

PR #1179, opened 2023-03-26
fix: quote killingTest in CSV report

### LINKED ISSUE #1178: Killingtest can contain commas, breaking CSV export
# Summary

The method definition of `killingTest` can contain commas, in case the method has parameters.
These commas break CSV parsing, as they are interpreted as starting a new column.

# Steps to reproduce

Simple example of code that will produce a breaking `killinTest`:

```java
@ParameterizedTest
@CsvSource({
    "test,test",
})
public void testMethod(String parm1, String param2 ) {
    // code left out
}
```
This would generate the following definition for `killingTest`

```testMethod(java.lang.String, java.lang.String)```

Notice the comma.

# What I expect to happen

That the method definition of the `killingTest` is quoted, as not to break CSV parsing.

# What actually happens

The `killingTest` is written to the CSV report as-is, breaking CSV parsing.

### PR DESCRIPTION
**Pull request by Bonajo**
This is a possible solution for #1178.  
It quotes the `killingTest` before writing to CSV report.

Fixes #1178

### DIFF
--- a/pitest-entry/src/main/java/org/pitest/mutationtest/report/csv/CSVReportListener.java
+++ b/pitest-entry/src/main/java/org/pitest/mutationtest/report/csv/CSVReportListener.java
@@ -37,7 +37,7 @@ public class CSVReportListener implements MutationResultListener {
   }
 
   private String createKillingTestDesc(final Optional<String> killingTest) {
-    return killingTest.orElse("none");
+    return (killingTest.isPresent()) ? String.format("\"%s\"", killingTest.get()) : "none";
   }
 
   private String makeCsv(final Object... os) {
--- a/pitest-entry/src/main/java/org/pitest/mutationtest/report/csv/CSVReportListener.java
+++ b/pitest-entry/src/main/java/org/pitest/mutationtest/report/csv/CSVReportListener.java
@@ -18,6 +18,8 @@ import java.io.IOException;
 import java.io.Writer;
 
 import java.util.Optional;
+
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by hcoles** [comment]
Thanks. This may now add the need to escape any double quotes in the text.

**Comment by Bonajo** [comment]
> Thanks. This may now add the need to escape any double quotes in the text.

As for as I understood quotes are not valid part of fully qualified names/method names, so it should not be possible for a quote to be in the `killingTest` definition. Or maybe I am missing something.

However if the CSV report would also include the description, such as the XML report, that would indeed be needed.

I am willing to work on this issue, if you have any ideas on how to improve, I more than happy to hear it.

**Comment by hcoles** [comment]
@Bonajo The PR is an improvement on the current situation for java users, so I'm happy to merge it as is as they make up the majority of the user base. However, tests written in kotlin could contain double quotes in their name. Escaping them would close off the possibility of this change causing issues in kotlin codebases.

**Comment by Bonajo** [comment]
I am not familiar with Kotlin, so I was unaware that these could contain double quotes. 

I noticed that `commons-text` is already a dependency, so I changed the quoting to use `escapeCsv` method from `StringEscapeUtils` instead.

This is a nicer solution and should work for both Java and Kotlin.

**Comment by szpak** [comment]
> However, tests written in kotlin could contain double quotes in their name.

The same with Groovy and Spock:

```
class FooSpec extends Specification {
    void "should pass"() {
        expect:
            1
    }
```

hcoles:
Thanks for your work on this @Bonajo

PR #1193, opened 2023-04-28
feat: ignore absent ReportAggregatorBuilder files

jrmcdonald:
### Summary
Similar to #466, when using the gradle-pitest-plugin it is possible to end up in a scenario where `linecoverage.xml` and `mutations.xml` file paths are sent to the `ReportAggregatorBuilder` but they do not actually exist.

An example of this would be where you have a multi module project with some nested sub modules:

```
sub-module/
    nested-module-a/
        src/
            main/
                ...
            test/
                ...
        build.gradle
    nested-module-b/
        src/
            main/
                ...
            test/
                ...
        build.gradle
build.gradle
```

The sub-module does not have a build.gradle, code, or tests, yet the plugin passes the file paths to the builder as if they did exist.
       
### Proposed Solution
I am proposing that the ReportAggregatorBuilder could log the missing files instead of raising an error. This would allow the aggregation to continue in cases such as the above, but still inform users of the missing output in case it was unexpected. 

I would welcome your thoughts on the above in case it has any unintended side effects.

### DIFF
--- a/pitest-aggregator/src/main/java/org/pitest/aggregate/ReportAggregator.java
+++ b/pitest-aggregator/src/main/java/org/pitest/aggregate/ReportAggregator.java
@@ -159,7 +159,11 @@ public final class ReportAggregator {
 
     public Builder addLineCoverageFile(final File lineCoverageFile) {
       validateFile(lineCoverageFile);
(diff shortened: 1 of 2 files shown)


jrmcdonald:
Awesome, thanks for accepting this!

PR #1194, opened 2023-05-02
Support glob pattern **. to match zero or more packages

### LINKED ISSUE #1144: Support for common glob pattern `**/` in arguments like `excludedClasses`
**Description**
It is very common to use `**/` in glob patterns to recursively match zero or more sub-directories. E.g. we want to reuse the exclude-Pattern List of Jacoco as input for `excludeClasses` including patterns like `**/databinding/**/*` but this doesn't match e.g. foo.databinding.Bar (of course replacing all `/` with `.`) in Pitest as Pitest doesn't cover the "zero sub-directories" case.

**Known Workaround**
Manually add multiple patterns, e.g. `*.databinding.*.*`, `*.databinding.*`, `databinding.*.*` and `databinding.*`

**Expected behavior**
Let `**.` match **zero or** more sub-directories.

Pfoerd:
Fixes #1144

Currently it is not possible to write a glob that matches zero or more packages. This PR introduces the well-kown doube-star matcher to enable this:

`**.databinding.**.Foo` 
* matches `databinding.Foo`
* matches `foo.car.databinding.bar.Foo`
* NOT matches `databindingfoo.Foo`

### DIFF
--- a/pitest/src/main/java/org/pitest/util/Glob.java
+++ b/pitest/src/main/java/org/pitest/util/Glob.java
@@ -25,6 +25,9 @@ public class Glob implements Predicate<String> {
 
   private final Pattern regex;
 
+  private static final String ZERO_OR_MORE_PACKAGES      = "(?:.*\\.)*";
+  private static final String DOUBLE_STAR_PACKAGE_MARKER = "#%#%#";
+
   public Glob(final String glob) {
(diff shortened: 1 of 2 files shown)


hcoles:
Thanks for this. Merged now, I'll push a release out shortly.

PR #1203, opened 2023-05-10
Fix JavaDoc of ArrayProbeCoverageMethodVisitor

Vampire:


Vampire:
Argh, was about to push an update as it was not correct yet. 🙈

hcoles:
@Vampire No worries, raise a new PR and it should get merged today.

I've not yet had chance to look at the concurrency related PRs, but just a heads up that if the motivation behind these is to support parallel running of junit 5 tests I am disinclined to try and support this. The simpler soloution would be to automatically disable parallel running. Supporting user questions on unexpected tests result is already difficuly and time consuming, adding concurrency into the mix will make maintainance harder.

Vampire:
@hcoles Well, the problem is you cannot automatically disable parallel running.
Yes, you did it by setting a system property for Jupiter.
But any other JUnit Platform Engine potentially has the same problem.
So you either have to support parallel running, or explicitly have Engine-specific code for all possible engines in the otherwise generic JUnit 5 plugin.
Actually, my ultimate goal was to support parallel running, yes, and the concurrency related PRs were also gearing towards this direction.
But I think even without fully supporting concurrent execution I think those changes make sense.

After I had a closer look at how the coverage recording is done, I also don't think it makes sense to support parallel execution, at least during coverage recording. To support it you would probably need to run each test with an own class loader or completely change the coverage recording which probably will not really work in any good way. And either solution would probably be slower in execution than to just prevent parallel execution.

During mutation testing it should imho work just fine to have the parallel execution working.

So currently one of my PRs at least prints a warning if it detects parallel test execution during coverage recording, and I'm playing with a way to generically prevent parallel test execution during coverage recording using locks instead of engine-specific code. That way coverage recording should work fine and sequentially while during mutant killing things could just run in parallel. And for that the concurrency related PRs I sent should also be relevant then. :-)

Issue #1182, opened 2023-03-29
why junit verstion be explicitly specififed, can the default junit version be not taken automatically


nagkumar:
Any new dependent property is extra, better pickup from gradle 

currently in 
```
pitest {
    junit5PluginVersion = '1.1.2'
    //pitestVersion = '1.11.7'
}
```

we have to give junit5PluginVersion , not giving this will results in 

> Task :pitest FAILED
8:54:50 am PIT >> INFO : Verbose logging is disabled. If you encounter a problem, please enable it before reporting an issue.
8:54:50 am PIT >> INFO : Incremental analysis reduced number of mutations by 0
8:54:50 am PIT >> INFO : Created 1 mutation test units in pre scan
8:54:50 am PIT >> INFO : Sending 2 test classes to minion
8:54:50 am PIT >> INFO : Sent tests to minion
8:54:50 am PIT >> SEVERE : Pitest could not run any tests. Please check that you have installed the pitest plugin for your testing library (eg JUnit 5, TestNG). If your project uses JUnit 4 the plugin is automatically included, but a recent version of JUnit 4 must be on the classpath.
Exception in thread "main" org.pitest.util.PitError: Please check you have correctly installed the pitest plugin for your project's test library (JUnit 5, TestNG, JUnit 4 etc).

nagkumar:
safe default configurations is the best way to make it more usage friendly.

hcoles:
This looks to be an issue with the gradle plugin. I suggest you raise this at https://github.com/szpak/gradle-pitest-plugin.

szpak:
@nagkumar It's not so trival - the plugin would need to know what pitest-junit5-plugin versions are compatible with used junit5(-platform) version. Nevertheless, you can crate a future requests in https://github.com/szpak/gradle-pitest-plugin/issues to discuss the possible options.

Btw, please appreciate that you have a quite clear error message - "missing pitest plugin for JUnit 5". Nevertheless, maybe this could be improved anyway.

nagkumar:
https://maven.apache.org/plugins/maven-toolchains-plugin/toolchains/custom.html 

can this approach of defining a custom toolchain help here.. just a thought.. not an expert on toolchains though

PR #1212, opened 2023-05-18
Auto add junit-platform-launcher to classpath

hcoles:
The pitest minions require a version of the junit-platform-launcher which matches the version of junit 5 used by the SUT. Previously this was provided by the cludge of shading it into the junit5 plugin, but this causes constant headaches with version mismatched.

This change pulls in a matching version of platform-launcher in the same manner that surefire does. It also allows junit dependencies to be explicitly included pitest config should this resolution give the wrong results.


nagkumar:
Is this solution avoiding the need for asking for an explicitly pitest-junit plugin version as asked at

https://github.com/hcoles/pitest/issues/1182

PR #1215, opened 2023-05-22
support code and coverage modifications when aggregating

hcoles:


hcoles:
@szpak I think the gradle plugin would need to be compiled against the next release to pick this up?

szpak:
> @szpak I think the gradle plugin would need to be compiled against the next release to pick this up?

In fact, I think that even not. Assuming the (`ReportAggregator`) API compatibility. the Gradle plugin fetches at runtime, the `pitest-aggregator`'s [version defined by an user](https://github.com/szpak/gradle-pitest-plugin/blob/c97fe8a7575674681badf5489712973effecd6a0/src/main/groovy/info/solidsoft/gradle/pitest/PitestAggregatorPlugin.groovy#L90-L96). It's good that you kept the constructor private, so that you needn't add the second one and using the build should handle that. As there doesn't seem to be any build related change to be implemented in the Gradle plugin and potential transformers are to be taken from PIT plugins available on the classpath, there is a change, it would just work only if an user points the new PIT version in the Gradle configuration.

Issue #681, opened 2019-09-26
inlined code detection warnings with try-with-resources


Vampire:
I have this code:
```java
Properties versionProperties = new Properties();
try (InputStream versionPropertiesStream = versionPropertiesResource.openStream()) {
    versionProperties.load(versionPropertiesStream);
} catch (IOException ignored) {
}
```
Line 81 is the line with the `catch`.

When I execute PIT with the inlined code detection enabled, I get four warnings about line 81, I added the mutation details via breakpoint:

```
09:45:53 PIT >> WARNING : Found more than one mutation similar on same line in a finally block. Can't correct for inlining.
MutationDetails [id=MutationIdentifier [location=Location [clazz=net.kautler.command.api.Version, method=<init>, methodDesc=()V], indexes=[66], mutator=org.pitest.mutationtest.engine.gregor.mutators.VoidMethodCallMutator], filename=Version.java, block=15, lineNumber=81, description=removed call to java/io/InputStream::close, testsInOrder=[], isInFinallyBlock=true, poison=NORMAL]
MutationDetails [id=MutationIdentifier [location=Location [clazz=net.kautler.command.api.Version, method=<init>, methodDesc=()V], indexes=[74], mutator=org.pitest.mutationtest.engine.gregor.mutators.VoidMethodCallMutator], filename=Version.java, block=17, lineNumber=81, description=removed call to java/lang/Throwable::addSuppressed, testsInOrder=[], isInFinallyBlock=true, poison=NORMAL]
MutationDetails [id=MutationIdentifier [location=Location [clazz=net.kautler.command.api.Version, method=<init>, methodDesc=()V], indexes=[79], mutator=org.pitest.mutationtest.engine.gregor.mutators.VoidMethodCallMutator], filename=Version.java, block=19, lineNumber=81, description=removed call to java/io/InputStream::close, testsInOrder=[], isInFinallyBlock=true, poison=NORMAL]
09:45:53 PIT >> WARNING : Found more than one mutation similar on same line in a finally block. Can't correct for inlining.
[...]

PR #1219, opened 2023-05-25
Filter try-with-resources before filtering inlined code again

Vampire:
Due to the changes in 1ceafb36, the issue in #681 that was fixed in #683 is back.
As the order of filters is no longer defined by the service loader file order,
but by name, the try-with-resources filter was put last and thus after the inlined code filter again.
This made the same issue reappear.

This PR adds an explicit order setting for features on which they are compared first and only for equal ones by name.
Standard value is 100 for all features, to prevent the psychologically bad necessity to use a negative order.
The try-with-resources filter gets 50, so it is executed before the inlined code filter which is the important part.


hcoles:
Good catch and good soloution. Thanks.

Vampire:
Thanks for adding the test. I was in a hurry and too lazy yesterday. :-D

Issue #521, opened 2018-09-30
Pitest 1.4.3 throws IllegalArgumentException


andrewbwogi:
Hello, I'm getting the following error when running Pitest 1.4.3 on the [DSpot test project](https://github.com/STAMP-project/dspot/tree/master/dspot/src/test/resources/test-projects) in Arch Linux 4.18.6 with Java 8:

```
Caused by: java.lang.IllegalArgumentException: Illegal base64 character 3c
    at java.util.Base64$Decoder.decode0 (Base64.java:714)
    at java.util.Base64$Decoder.decode (Base64.java:526)
    at java.util.Base64$Decoder.decode (Base64.java:549)
    at org.pitest.mutationtest.incremental.ObjectOutputStreamHistoryStore.deserialize (ObjectOutputStreamHistoryStore.java:126)
    at org.pitest.mutationtest.incremental.ObjectOutputStreamHistoryStore.restoreClassPath (ObjectOutputStreamHistoryStore.java:114)
    at org.pitest.mutationtest.incremental.ObjectOutputStreamHistoryStore.initialize (ObjectOutputStreamHistoryStore.java:86)
    at org.pitest.mutationtest.tooling.MutationCoverage.runReport (MutationCoverage.java:132)
    at org.pitest.mutationtest.tooling.EntryPoint.execute (EntryPoint.java:120)
    at org.pitest.mutationtest.tooling.EntryPoint.execute (EntryPoint.java:50)
    at org.pitest.maven.RunPitStrategy.execute (RunPitStrategy.java:35)
    at org.pitest.maven.AbstractPitMojo.analyse (AbstractPitMojo.java:455)
    at org.pitest.maven.AbstractPitMojo.execute (AbstractPitMojo.java:396)
    at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo (DefaultBuildPluginManager.java:137)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:208)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:154)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:146)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:117)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:81)
[...]

PR #1261, opened 2023-10-10
do not error when given old history file

hcoles:


jtymes:
Thank you @hcoles. Does this close #521?

PR #1308, opened 2024-02-08
bump mvn vcs versions

hcoles:


milindbangar79:
Apologies Henry , for not responding . Been tied up with quite a few things . I will check and revert by Monday

PR #1310, opened 2024-02-12
filter mutations to defensive unmodifiable collection returns

hcoles:
It's common practice in some teams to wrap mutable collections with unmodifiable wrappers before returning them from a class to prevent accidental modification of state.

```java
class Foo {
   private final List<String> field;

   ...

   List<String> doStuff() {
     return Collections.unmodifiableList(field);
   }
}
```

Although the presence of the wrapper can be easily detected by a test that attempts to modify the state of the list, some teams do not wish to specify this behaviour.

This change provides the option to filter out mutations implementing this defensive coding pattern.

The filter is disabled by default, but can be enabled with the `+funmodifiablecollection` feature string.


romani:
@hcoles , is this a way to activate this filter https://pitest.org/quickstart/maven/#features ?

PR #1353, opened 2024-10-04
warn about goal removal

hcoles:


brunoduarte40:
I use it a lot and I haven't had any problems with maintenance. What was the maintenance headaches?

CaioCesar-cloud:
I use it a lot, and I think it's a viable way to correct it in my team, as it has better performance, as I use it in a large project, when using SCM I save time, what are the problems?

hcoles:
@brunoduarte40 The maintenance problem is that the SCM goal has to be considered when making any changes to the maven plugin and consideration has to be given to how it interacts with planned new features.

CaioCesar-cloud:
And do you suggest any alternatives to use that are similar to SCM?

hcoles:
If you're using git, https://www.arcmutate.com provides a much better integration.

lennart-bos:
> @brunoduarte40 The maintenance problem is that the SCM goal has to be considered when making any changes to the maven plugin and consideration has to be given to how it interacts with planned new features.
What planned new features? Will there be a replacement similar to the SCM goal?

Issue #1360, opened 2024-11-03
How do I simply exclude method from class


see-quick:
Hello, how can one configure within the JUnit5 plugin exclusion of the specific method? Imagine simple example:
```java
package io;

import org.junit.jupiter.api.Tag;

import java.util.List;
import java.util.stream.Collectors;

public class MyClass {

    /**
     * Filters a list of strings, returning only those that start with a specified prefix.
     *
     * @param input  the list of strings to filter
     * @param prefix the prefix to filter by
     * @return a list of strings that start with the given prefix
     */
    public List<String> filterByPrefix(List<String> input, String prefix) {
        return input.stream()
            .filter(s -> s.startsWith(prefix))
            .collect(Collectors.toList());
    }
}
```
and dummy tests, which does nothing (expected) because we want to exclude such method.
```
package io;

import org.junit.jupiter.api.Test;


public class MyClassTest {

    @Test
    public void testFilterByPrefix() {
        // Dummy nothing
    }
}
```
Now when I run Pitest with the default configuration I get 
<img width="1030" alt="image" src="https://github.com/user-attachments/assets/1ae60b68-7384-465c-95f4-9d4396ddd050">

which is fine because I didn't exclude any method yet. But now when I add to the plugin configuration 
```
 <excludedMethods>
        <param>io.MyClass::lambda$filterByPrefix$*</param>
</excludedMethods>
```

nothing really changed. I tried various stuff:
1. around <param> without `/` with `/
2. escaped $ 
3. etc.

My question is that possible to simply exclude a method? Because this is really great use case when you have a class where logic is basically **mainly** tasted on the Integration level and not the Unit level. I tried the same with the <excludedClasses> this works well without any problems. Thanks for the quick answer.

see-quick:
> The excludedMethods parameter is designed to exclude all methods matching a given glob. It does not match against the class name containing it, so "io.MyClass::" guarantees nothing will ever match.
> 
> To exclude specific methods you have three options.
> 
> 1. Annotate them with an annotation named Generated, CoverageIgnore, or DoNotMutate with at least class level retention
> 2. Create a [custom mutation filter](https://pitest.org/quickstart/advanced/)
> 3. Use the [arcmutate base exclusion extensions](https://docs.arcmutate.com/docs/exclusions.html)

> 1. Annotate them with an annotation named Generated, CoverageIgnore, or DoNotMutate with at least class level retention

[...]

PR #1362, opened 2024-11-06
Mutation coverage fix within lambdas 

see-quick:
This PR fixes the problem described [1]. 

[1] - https://github.com/hcoles/pitest/issues/1360

### DIFF
--- a/pitest-entry/src/main/java/org/pitest/mutationtest/build/intercept/annotations/ExcludedAnnotationInterceptor.java
+++ b/pitest-entry/src/main/java/org/pitest/mutationtest/build/intercept/annotations/ExcludedAnnotationInterceptor.java
@@ -42,7 +42,6 @@ public class ExcludedAnnotationInterceptor implements MutationInterceptor {
           .filter(hasAvoidedAnnotation())
           .map(AnalysisFunctions.matchMutationsInMethod())
           .collect(Collectors.toList());
-
       this.annotatedMethodMatcher = Prelude.or(methods);
     }
   }
--- a/pitest-entry/src/main/java/org/pitest/mutationtest/build/intercept/annotations/ExcludedAnnotationInterceptor.java
+++ b/pitest-entry/src/main/java/org/pitest/mutationtest/build/intercept/annotations/ExcludedAnnotationInterceptor.java
@@ -24,7 +24,6 @@ public class ExcludedAnnotationInterceptor implements MutationInterceptor {
   private boolean skipClass;
   private Predicate<MutationDetails> annotatedMethodMatcher;
 
-
   ExcludedAnnotationInterceptor(List<String> configuredAnnotations) {
     this.configuredAnnotations = configuredAnnotations;
   }
@@ -43,13 +42,15 @@ public class ExcludedAnnotationInterceptor implements MutationInterceptor {
           .filter(hasAvoidedAnnotation())
           .map(AnalysisFunctions.matchMutationsInMethod())
           .collect(Collectors.toList());
+
       this.annotatedMethodMatcher = Prelude.or(methods);
     }
   }
 
   private Predicate<MethodTree> hasAvoidedAnnotation() {
-    return a -> a.annotations().stream()
-        .anyMatch(avoidedAnnotation());
+    return methodTree ->
+        // count also lambda generated methods
+        methodTree.isGeneratedLambdaMethod() || methodTree.annotations().stream().anyMatch(avoidedAnnotation());
   }
 
   private Predicate<AnnotationNode> avoidedAnnotation() {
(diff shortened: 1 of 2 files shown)


hcoles:
This looks like it will remove mutations defined in all lambdas,  not just lambdas defined within the an annotated method.

see-quick:
> This looks like it will remove mutations defined in all lambdas, not just lambdas defined within the an annotated method. This is why a test is failing in the overall suite.
> 
> I think a minimum set of test cases would be
> 
> * unannotated method
> * annotated method
> * lambda within unannotated method
> * lambda within annotated method
> * lambda within unannotated overridden method
> * lambda within annotated overridden method
> * lambda within lambda within unannotated overridden method
> * lambda within lambda within annotated overridden method

Yeah I will add those tests. Thanks for pointing it out.

pitest-entry/src/main/java/org/pitest/mutationtest/build/intercept/annotations/ExcludedAnnotationInterceptor.java
```diff
@@ -39,17 +40,56 @@ public void begin(ClassTree clazz) {
...
+      final Set<String> avoidedMethodNames = avoidedMethods.stream()
+          .map(method -> method.rawNode().name)
+          .collect(Collectors.toSet());
+
+      // 2. Collect lambda methods with being inside avoided methods
+      final List<MethodTree> lambdaMethods = clazz.methods().stream()
+          .filter(MethodTree::isGeneratedLambdaMethod)
+          .filter(lambdaMethod -> {
+            String lambdaName = lambdaMethod.rawNode().name; // e.g., lambda$fooWithLambdas$0
+            String enclosingMethodName = extractEnclosingMethodName(lambdaName);
```
hcoles, line 57: The logic here is based solely on method name. If you were to add the other suggested tests they would fail as all lambdas in all the overloads of a method would be filtered if any one of the overloads was annotated.

see-quick: I have updated the tests. Also, the logic should be adjusted to overridden methods. Let me know if there's something I miss...Thanks

see-quick: Hmm interesting that the build is failing after 1m but locally my build is successful without any problems. 
```java
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for pitest-parent dev-SNAPSHOT:
[INFO] 
[INFO] pitest-parent ...................................... SUCCESS [  0.365 s]
[INFO] Build Config ....................................... SUCCESS [  0.295 s]
[INFO] pitest ............................................. SUCCESS [  6.108 s]
[INFO] pitest-entry ....................................... SUCCESS [ 51.068 s]
[INFO] pitest-html-report ................................. SUCCESS [  0.782 s]
[INFO] pitest-aggregator .................................. SUCCESS [  0.795 s]
[INFO] pitest-maven ....................................... SUCCESS [  5.872 s]
[INFO] pitest-command-line ................................ SUCCESS [  0.637 s]
[INFO] pitest-ant ......................................... SUCCESS [  0.626 s]
[INFO] pitest-java8-verification .......................... SUCCESS [  3.993 s]
[INFO] pitest-maven-verification .......................... SUCCESS [02:09 min]
[INFO] pitest-modern-verification ......................... SUCCESS [  0.670 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  03:20 min
[INFO] Finished at: 2024-11-08T10:39:43+01:00
[INFO] ------------------------------------------------------------------------
```

hcoles: I don't think overrides need to be considered, just overloads within the same class. Although it is possible to navigate the class hierarchy from the bytecode, it becomes quite expensive when there are a lot of mutants.

Unhelpfully, I used the word override when I meant overload earlier when describing the test cases.

The failing test is likely because the suite runs a matrix of different java versions. Sometimes the bytecode generated differs substantially with version. Tests need to be careful what they assert on in case they are making implicit assumptions.

see-quick: Updated for overloads. Tested also with my code and it seems to work fine. Build locally is passing now I have to figure it out to fix it also upstream :D

see-quick:
It seems I have fixed everything. Could you look at that when you have some free time @hcoles? Many thanks.

PR #1369, opened 2024-12-13
Feature/dry run support

hcoles:


romani:
@hcoles , is there user friendly explanation of this feature?
In what cases user get benefits of this mode?

hcoles:
@romani It's explained here under Dry Run Mode (it appears I've still not gotten round to add link anchors to the subtitles)

https://pitest.org/quickstart/maven/

Basically it lets you run tests and generate mutants without going on to analyse the mutants (i.e do the expensive bits).

This is useful when first setting pitest up for a project if you're having issues with failing tests, handy for me when investigating issues as it gives me an easy way to generate mutants and examine their bytecode, and will shortly also become handy for arcmutate users. 

I'm working an enhancement to the arcmutate history implementation

https://docs.arcmutate.com/docs/history.html

Once complete, it will be able to use the limited information gather during the dry run to accelerate full analyses. This will be handy for some of our customers who have large monolithic codebases where the tests take a long time to run, even before mutation analysis begins.

PR #1387, opened 2025-02-17
Quote all classpath entries to handle spaces

hcoles:


pitest/src/test/java/org/pitest/classpath/CompoundClassPathRootTest.java
```diff
@@ -11,6 +11,7 @@
 import java.net.URL;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.Optional;
```
szpak, line 14: ?

pitest-entry/src/main/java/org/pitest/process/Java9Process.java
```diff
@@ -108,8 +108,8 @@ private List<String> createLaunchArgs(JavaAgent agentJarLocator, List<String> ar
         List<String> cmd = new ArrayList<>();
 
         cmd.add("-classpath");
-        cmd.add(classPath);
-
+        cmd.add(classPath.replace(" ", "\" \""));
```
szpak, line 111: I wonder, if it can affect the Gradle plugin, in the way how it already passes the classpath entries :thinking:

hcoles: It has just occurred to me that this change would cause a problem if gradle were already quoting the the spaces. Otherwise, this should have no impact.

PR #1424, opened 2025-09-10
Auto-configure JUnit Jupiter dependencies

rycus86:
In our large multi-module monorepo we found ourselves having to add the `junit-jupiter-engine` dependency to modules so that pitest can run mutation tests, but ideally we'd love to just add the engine dependency to the pitest plugin execution instead as Surefire/Failsafe doesn't need it.

Our current workaround is to add the engine dependency _and_ configure the additional classpath, but it would be nice if that happened automatically (like it does for the platform jars already).

```xml
<additionalClasspathElements>
    ${settings.localRepository}/org/junit/jupiter/junit-jupiter-engine/${junit5.version}/junit-jupiter-engine-${junit5.version}.jar
</additionalClasspathElements>
```

### DIFF
--- a/pitest-maven/src/main/java/org/pitest/maven/MojoToReportOptionsConverter.java
+++ b/pitest-maven/src/main/java/org/pitest/maven/MojoToReportOptionsConverter.java
@@ -417,9 +417,12 @@ public class MojoToReportOptionsConverter {
       classPath.add(dependency.getFile().getAbsolutePath());
     }
 
-    // If the user as explicitly added junit platform classes to the pitest classpath, trust that they
-    // know what they're doing and add them in
-    this.mojo.getPluginArtifactMap().values().stream().filter(a -> a.getGroupId().equals("org.junit.platform"))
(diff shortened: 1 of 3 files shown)


rycus86:
Thanks for having a look @hcoles !
Yes, you're right this is a workaround and I think https://github.com/hcoles/pitest/pull/1426 should fix what I wanted, but I'll try and verify today. Thank you!

rycus86:
Yes, I can confirm it works with the new version, thank you!
```
[INFO] Adding org.pitest:pitest-junit5-plugin to SUT classpath
[INFO] Adding org.pitest:pitest to SUT classpath
[INFO] Auto adding org.junit.platform:junit-platform-launcher:jar:1.12.2 < ... to classpath.
[INFO] Auto adding org.junit.platform:junit-platform-engine:jar:1.12.2 < ... to classpath.
[INFO] Auto adding org.junit.jupiter:junit-jupiter-engine:jar:5.12.2 < ... to classpath.
```

PR #1455, opened 2026-02-04
move default history analysis to plugin

hcoles:
Moves incremental analysis to a seperate plugin at https://github.com/pitest/pitest-history


szpak:
Is the new plugin a drop-in replacement (once upgraded to the newest version + the plugin added as a dependency)?

hcoles:
@szpak Yes, it's the same code, but packaged as an external plugin. The hope is to find a new maintainer for it once it's separated from the main codebase.

PR #1461, opened 2026-04-04
fix(Java9Process): remove stale args file cache

tejasae-afk:
The `CACHE` in `Java9Process` keys on `javaAgent.getJarLocation()` but caches the full args file — which includes classpath, JVM options, and mutation targets that differ between minion invocations. A second minion launched with a different classpath or config silently reuses the first minion's args file and runs with the wrong arguments.

Removing the cache is the safe fix. `createArgsFile()` writes a small temp file and isn't on the hot path — the cache was not buying anything meaningful here.

Happy to close this if there's a reason for the cache I'm not seeing — you know this codebase far better than I do.

### DIFF
--- a/pitest-entry/src/main/java/org/pitest/process/Java9Process.java
+++ b/pitest-entry/src/main/java/org/pitest/process/Java9Process.java
@@ -9,9 +9,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Map;
 import java.util.Optional;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Predicate;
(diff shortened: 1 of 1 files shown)


hcoles:
The code is admittedly very unclear but I believe functions correctly. The args should always be the same for a given agent location within a run so can safely be used a key. The cache may or may not have a measurable performance impact, but it does prevent large numbers of temporary files from being written to the filesystem.

Are you observing some behaviour that suggests there is a bug?

tejasae-afk:
Thanks for the clarification — that context makes the cache make sense. I was looking at the code in isolation and assumed fileArgs could vary across calls with the same agent, but if the agent location uniquely determines the args within a run that changes things. I haven't observed actual broken behavior — this was a static analysis finding, not a reproduction.

Issue #1464, opened 2026-04-28
Support configurable decimal precision for coverage and mutation thresholds


see-quick:
When I was re-searching about this topic I found out multiple related issues i.e. https://github.com/hcoles/pitest/issues/372,  https://github.com/hcoles/pitest/issues/784 or https://github.com/hcoles/pitest/issues/140.

But nothing really solved line coverage issue (only mutation coverage with `maxSurviving` configuration).

### Problem

The `coverageThreshold`, `mutationThreshold`, and `testStrengthThreshold` parameters only accept integer percentages. This is really not so precise as it creates a significant blind spot where code changes shift the denominator, causing the percentage to fluctuate without any meaningful change in test quality (i.e., denominator shifting)

For instance let's assume I have this report: 

| Classes | Line Coverage       | Mutation Coverage  | Test Strength      |
|---------|---------------------|--------------------|--------------------|
| 9       | 61% (796/1295)      | 78% (324/413)      | 97% (324/335)      |

The actual line coverage is  **61.47%**. With `<coverageThreshold>61</coverageThreshold>`:

| Change                          | Actual % | Rounds to | Result                                   |
|---------------------------------|----------|-----------|------------------------------------------|
| Lose 12 covered lines (796 to 784) | 60.54%   | 61%       | **passes** silent regression           |
| Add 20 untested lines           | 60.53%   | 61%       | **passes** no tests, no problem        |
| Add 21 untested lines           | 60.49%   | 60%       | **fails** 1 line difference from above |
| Add 1 covered line              | 61.54%   | 62%       | jumps a whole percent for 1 line         |

With integer thresholds, the blind spot is **~1 full percentage point**. 
So basically with project where you have 1000 lines of code the blind spot is 10 lines of coverage that can seilently drift without the threshold noticing. You can imagine how blind spot is really huge when you for instance check codebase arround 100k LOC (i.e., it's 1000 lines)

### Proposal

I would like to propose a new `thresholdPrecision` parameter that controls how many decimal places are used when computing and comparing coverage/mutation percentages. Default is `0` (current behavior), making this fully backward compatible.

So eventually users would choose level of granularity that makes sense for their project. As I can imagine as small projects might be fine with using 1 or 2, larger ones could have even 7 and more.  

If you think it make sense to implement I would be happy to cotribute. @hcoles.

### Rejected

[...]

PR #1468, opened 2026-05-16
Add support for configurable decimal precision

see-quick:
This change adds a `thresholdPrecision` parameter (default 0, fully backward compatible) that controls how many decimal places are used when computing and comparing coverage and mutation thresholds. 
With integer only thresholds, there is an unavoidable `~1` percentage point blind spot where coverage can silently regress without triggering a build failure.   
  
For example, a project with coverageThreshold=61 won't catch a drop from 61.8% to 61.1% (i.e., both round to 61%).                                                                    
                                                                                                                                                                                     
Moreover, rather than maintaining two parallel APIs (i.e., one returning `int`, another `BigDecimal`) the thresholds were changed to BigDecimal throughout. A single path with `precision` set to `0` produces identical results to the old integer behaviour, so the dual-API approach would add complexity without any behaviour benefit.

Closing #1464. 

I have tested local build with my project, and it seems to work fine without any issues. Let me know what you think about it @hcoles. Hopefully I didn't forgot anything. 

This is project view I had to make a few CSS/HTLM tweaks but eventually get it worked. 
#### Project view: 

<img width="995" height="360" alt="project_view" src="https://github.com/user-attachments/assets/8bc2c309-c48b-4f34-be33-a749c896f11c" />

#### Package view:
<img width="975" height="374" alt="package_view" src="https://github.com/user-attachments/assets/13d3ac12-bcd5-479a-96d8-560bf8afdf52" />

### DIFF
--- a/pitest-aggregator/src/main/java/org/pitest/aggregate/AggregationResult.java
+++ b/pitest-aggregator/src/main/java/org/pitest/aggregate/AggregationResult.java
@@ -1,13 +1,14 @@
 package org.pitest.aggregate;
 
+import java.math.BigDecimal;
+
 public class AggregationResult {
   private final long mutations;
   private final long mutationsSurvived;
-  private final int  mutationCoverage;
+  private final BigDecimal mutationCoverage;
+  private final BigDecimal testStrength;
 
-  private final int testStrength;
-
-  public AggregationResult(long mutations, long mutationsSurvived, int mutationCoverage, int testStrength) {
+  public AggregationResult(long mutations, long mutationsSurvived, BigDecimal mutationCoverage, BigDecimal testStrength) {
     this.mutations = mutations;
     this.mutationsSurvived = mutationsSurvived;
     this.mutationCoverage = mutationCoverage;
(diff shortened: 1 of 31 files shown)


see-quick:
Gentle ping @hcoles.

hcoles:
@see-quick I should be able to look at this next week.

see-quick:
Hi @hcoles, just checking in. Any updates on this?

hcoles:
Merged now, thanks.

If you could also raise a PR against

https://github.com/hcoles/pitest-site

To document the changes, that would be helpful.

see-quick:
> Merged now, thanks.
> 
> If you could also raise a PR against
> 
> https://github.com/hcoles/pitest-site
> 
> To document the changes, that would be helpful.

Yeah, I will try this week, if I find some free cycles. Thanks.

see-quick:
Do you think there could be a patch release (i.e., 1.25.4) for this soon?

hcoles:
@see-quick - released now.

PR #1469, opened 2026-05-20
Progress report

eatdrinksleepcode:
The mutation analysis phase can take a long time, but there's no indication of how far along it is. This adds periodic progress logging that reports how many mutation test units have completed out of the total, giving a rough sense of how much time remains.

- Adds a scheduled progress reporter to `MutationAnalysisExecutor` that logs completion count at a configurable interval                                                                                       
- Exposes `progressInterval` (in seconds) as a configuration option via CLI and Maven plugin                                                                                                                 
- Defaults to disabled (0) for backwards compatibility; opt in by setting a positive interval

### DIFF
--- a/pitest-entry/src/main/java/org/pitest/mutationtest/config/SettingsFactory.java
+++ b/pitest-entry/src/main/java/org/pitest/mutationtest/config/SettingsFactory.java
@@ -267,7 +267,8 @@ public class SettingsFactory {
 
   private static Predicate<MutationResultListenerFactory> matchesOutputFormat(
           final Iterable<String> outputFormats) {
-    return a -> FCollection.contains(outputFormats, equalsIgnoreCase(a.name()));
+    return a -> a.provides().equals(MutationResultListenerFactory.LEGACY_MODE)
+            && FCollection.contains(outputFormats, equalsIgnoreCase(a.name()));
   }
 
   private static Predicate<MutationResultListenerFactory> isActivatedByFeature() {
--- a/pitest-entry/src/main/java/org/pitest/mutationtest/config/SettingsFactory.java
+++ b/pitest-entry/src/main/java/org/pitest/mutationtest/config/SettingsFactory.java
@@ -249,21 +249,29 @@ public class SettingsFactory {
   private Collection<MutationResultListenerFactory> findListeners() {
     final Iterable<? extends MutationResultListenerFactory> listeners = this.plugins
         .findListeners();
-    final Collection<MutationResultListenerFactory> matches = FCollection
-        .filter(listeners, nameMatches(this.options.getOutputFormats()));
-    if (matches.size() < this.options.getOutputFormats().size()) {
+
+    final Collection<MutationResultListenerFactory> outputFormatListeners = FCollection
+        .filter(listeners, matchesOutputFormat(this.options.getOutputFormats()));
+    if (outputFormatListeners.size() < this.options.getOutputFormats().size()) {
       throw new PitError("Unknown listener requested in "
           + String.join(",", this.options.getOutputFormats()));
     }
-    return matches;
+
+    final Collection<MutationResultListenerFactory> featureListeners = FCollection
+        .filter(listeners, isActivatedByFeature());
+
+    final Collection<MutationResultListenerFactory> all = new ArrayList<>(outputFormatListeners);
+    all.addAll(featureListeners);
+    return all;
   }
 
-  private static Predicate<MutationResultListenerFactory> nameMatches(
+  private static Predicate<MutationResultListenerFactory> matchesOutputFormat(
           final Iterable<String> outputFormats) {
-    // plugins can be either activated here by name
-    // or later via the feature mechanism
-    return a -> FCollection.contains(outputFormats, equalsIgnoreCase(a.name()))
-            || !a.provides().equals(MutationResultListenerFactory.LEGACY_MODE);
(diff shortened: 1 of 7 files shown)


hcoles:
Thanks for this.

I think it should be possible to implement this as a org.pitest.mutationtest.MutationResultListener.

This would require far less code change, and would automatically make it a configurable feature across all build tools, including gradle.

The org.pitest.mutationtest.MutationResultListenerFactory allow a feature string to be configured, so this could be turned on with, for example 

+progress.

Features can also accept simple parameters, so the interval could be set with

+progress(interval[42])

The examples of MutationResultListenerFactory in the codebase don't demonstrate how the feature strings work (features were retrofitted to the interface for the benefit of external plugins). But if you look at the MutationInterceptorFactory there are several examples.

eatdrinksleepcode:
Funny enough, after I implemented this I wanted to test it with my existing Gradle projects, and I realized the Gradle plugin only allows known configuration options. I then figured out that it _does_ allow specifying arbitrary features, so I implemented a `ConfigurationUpdater` that works exactly like you mentioned: `+progress(interval[42])`. But it felt like a workaround for a Gradle plugin limitation, so I kept that part local and did not include it.

I'll have a look at the `MutationResultListener` approach.

eatdrinksleepcode:
@hcoles The downside of the `MutationResultListener` approach is that we lose out on the denominator. `MutationAnalysisExecutor` knows the total number of mutation units before it starts, so although all units are not equal in size, the progress report has a useful percentage component. With `MutationResultListener` we could only report that absolute progress is being made. That helps show that the analysis is not stalled, but it doesn't give the same sense of percentage progress that I was going for.

One solution would be to add the total number of mutation units to the `ListenerArguments` data structure that is passed to `MutationResultListenerFactory.getListener`. The total has already been calculated before `ListenerArguments`.

Another option would be to expose more information about the scan on the `ClassMutationResults` instance that is passed to `MutationResultListener`; either as individual properties, or as a `MutationAnalysisInfo` object that has information about the scan, or even a `MutationAnalysisResults` that aggregates the results so far. It's not the worst approach, but feels a bit hackish.

pitest-entry/src/main/java/org/pitest/mutationtest/config/SettingsFactory.java
```diff
@@ -257,21 +249,30 @@ public CoverageTransformer createCoverageTransformer(CodeSource code) {
   private Collection<MutationResultListenerFactory> findListeners() {
     final Iterable<? extends MutationResultListenerFactory> listeners = this.plugins
         .findListeners();
-    final Collection<MutationResultListenerFactory> matches = FCollection
-        .filter(listeners, nameMatches(this.options.getOutputFormats()));
-    if (matches.size() < this.options.getOutputFormats().size()) {
+
+    final Collection<MutationResultListenerFactory> outputFormatListeners = FCollection
+        .filter(listeners, matchesOutputFormat(this.options.getOutputFormats()));
+    if (outputFormatListeners.size() < this.options.getOutputFormats().size()) {
```
eatdrinksleepcode, line 255: Because there is now a `MutationResultListenerFactory` that does not use the `LEGACY_MODE` feature, this guard did not fire correctly when an invalid output format was specified, because there was 1 requested output format and 1 listener in the resulting collection (it was just the wrong listener). The corresponding test in `SettingsFactoryTest` caught the problem. Output formats are now verified against `LEGACY_MODE` listeners _only_; then feature-managed listeners are added in.

pitest-entry/src/test/java/org/pitest/mutationtest/report/progress/ProgressListenerFactoryTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class ProgressListenerFactoryTest {
```
eatdrinksleepcode, line 19: Other listener factories don't have tests, but I felt like these are simple enough and have enough value to add. Can remove if you don't think they are worth keeping.

pitest-entry/src/test/java/org/pitest/mutationtest/report/progress/ProgressListenerTest.java
```diff
@@ -0,0 +1,45 @@
...
+    public void shouldCountIndividualMutationsAcrossMultipleResults() throws Exception {
+        ProgressListener testee = new ProgressListener(out, 1);
+        testee.runStart();
+        try {
+            testee.handleMutationResult(createClassResults(aMutationTestResult().build()));
+            testee.handleMutationResult(createClassResults(
+                    aMutationTestResult().build(),
+                    aMutationTestResult().build(),
+                    aMutationTestResult().build()));
+            Thread.sleep(1500);
```
eatdrinksleepcode, line 28: I don't love that this test has to sleep for a 1.5 seconds. One alternative is to express the interval in `ProgressListener` in milliseconds, and convert the values in `ProgressListenerFactory`. Then at least this test could sleep for a much shorter time. That's actually what I did in the previous version of this PR. But ultimately I decided the confusion of expressing the value with different units in different places wasn't worth it.

eatdrinksleepcode (commented):
@hcoles I updated the implementation using a `MutationResultListener` as you suggested. It's definitely cleaner and does not require modifying core classes. The only downside is the loss of the denominator.

